PR #13, opened 2016-07-19
Create gh pages

gforsyth:
Adding in the automatic creation of a gh-pages branch if it doesn't exist already on the remote

### DIFF
--- a/travis_docs_builder.py
+++ b/travis_docs_builder.py
@@ -174,6 +174,38 @@ def setup_GitHub_push(repo):
 
     return True
 
+def gh_pages_exists():
+    """
+    Check if there is a remote gh-pages branch.
+    This isn't completely robust.  If there are multiple remotes and you have a gh-pages
+    branch on the non-default remote, this won't see it.
+    """
+    remote_name = subprocess.check_output(['git','remote', 'show']).decode('utf-8').split()[0]
+    branch_names = subprocess.check_output(['git', 'branch', '-r']).decode('utf-8').split()
+
+    return '{}/gh-pages'.format(remote_name) in branch_names
+
+def create_gh_pages():
+    """
+    If there is no remote gh-pages branch, create one
+
+    Return True if gh-pages was created, False if not
+    """
+    if not gh_pages_exists():
+        print("Stashing your current unstaged work")
+        run(['git', 'stash'])
+        print("Creating gh-pages branch")
+        run(['git', 'symbolic-ref', 'HEAD', 'refs/heads/gh-pages'])
+        #delete everything in the new ref.  this is non-destructive to existing
+        #refs/branches, etc...
+        run(['git', 'rm', '-rf', '.'])
+        print("Adding .nojekyll file to gh-pages branch")
+        run(['touch', '.nojekyll'])
+        run(['git', 'add', '.nojekyll'])
+        run(['git', 'commit', '-m', '"create new gh-pages branch with .nojekyll"'])
+
+        return True
+    return False
 
 # Here is the logic to get the Travis job number, to only run commit_docs in
 # the right build.
--- a/travis_docs_builder.py
+++ b/travis_docs_builder.py
@@ -167,8 +167,9 @@ def setup_GitHub_push(repo):
         'https://{token}@github.com/{repo}.git'.format(token=token.decode('utf-8'), repo=repo)])
     print("Fetching token remote")
     run(['git', 'fetch', 'origin_token'])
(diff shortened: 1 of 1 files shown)


asmeurer:
Ah, it seems we also need to implement the "don't push on pull requests" feature :)

travis_docs_builder.py
```diff
@@ -167,13 +167,48 @@ def setup_GitHub_push(repo):
...
+
+def create_gh_pages():
+    """
+    If there is no remote gh-pages branch, create one
+
+    Return True if gh-pages was created, False if not
+    """
+    if not gh_pages_exists():
+        print("Creating gh-pages branch")
+        run(['git', 'symbolic-ref', 'HEAD', 'refs/heads/gh-pages'])
```
asmeurer: Does this create an empty branch?

gforsyth: not by itself, no.  but you create the new ref, then nuke everything, then add `.nojekyll` and then commit and end up with a branch that has only the `.nojekyll` to start with

asmeurer:
Would be nice to test this. Maybe for now we could just wipe the `gh-pages` branch on this repo to test it. 

That reminds me, we should enable branch protection for `master` on this repo so we don't accidentally push to it with our shenanigans. You can make it so that all changes have to come in through PRs, which is a little annoying, but I think worth it to avoid the risk, since we are literally working on a script that pushes to GitHub.

gforsyth:
Ok, branch protection enabled for `master`.  I'll delete the gh-pages branch from here and then restart the travis build

asmeurer:
I wonder how to test this more generally. A dummy testing repo on GitHub that somehow gets run with pushes for this thing? There's some security implications there, since this thing scopes all public repos. Sometimes writing proper tests for a thing is harder than writing the thing itself.

travis_docs_builder.py
asmeurer: Right, but it doesn't have history. That's the important thing. I just checked in a test repo, and this seems to do what we want.

asmeurer: Supposedly you can do `git checkout --orphan` as well http://stackoverflow.com/a/13969482/161801

gforsyth:
Yeah, that's a tricky one.  Also, I realize now that this won't test the gh-pages behavior until we merge it, since it won't run on this branch

travis_docs_builder.py
gforsyth: hmm, i learned something new.  any preference on method?

asmeurer: I would use the checkout version. `symbolic-ref` seems like a plumbing command, but it's better to use porcelain commands to do high-level things, if possible.

asmeurer: Then again, for programmatic access you are supposed to use plumbing commands (https://stackoverflow.com/questions/2978947/how-to-list-all-the-log-for-current-project-in-git/2979124#comment3041215_2979124, https://stackoverflow.com/questions/3878624/how-do-i-programmatically-determine-if-there-are-uncommited-changes/3879077#3879077). So I don't know.

gforsyth: Ha.  well, I just changed it to `orphan` and it seems a little cleaner to me.  i'm good with that

asmeurer:
Seems like it should work. Feel free to merge/test it if you feel the same.

asmeurer:
Looks like it didn't work https://travis-ci.org/gforsyth/travis_docs_builder/builds/146468700

travis_docs_builder.py
```diff
@@ -167,13 +167,48 @@ def setup_GitHub_push(repo):
...
+    If there is no remote gh-pages branch, create one
+
+    Return True if gh-pages was created, False if not
+    """
+    if not gh_pages_exists():
+        print("Creating gh-pages branch")
+        run(['git', 'checkout', '--orphan', 'gh-pages'])
+        #delete everything in the new ref.  this is non-destructive to existing
+        #refs/branches, etc...
+        run(['git', 'rm', '--cached', '-rf', '.'])
```
asmeurer: I guess this untracks the files, but doesn't actually delete them. Maybe should just do a `git reset --hard`?

asmeurer: Oh that's specifically what the `--cached` option does:

```
       --cached
           Use this option to unstage and remove paths only from the index. Working tree files, whether modified or not, will be left alone.
```

(from `git help rm`)

gforsyth: dammit.  that was silly.  fix incoming

PR #14, opened 2016-07-21
Don't push docs on pull requests or non-master

asmeurer:
It shouldn't push for the PR build. I also pushed this branch up directly, to test that.


gforsyth:
Nice! this seems solid to me.

asmeurer:
Yep, you can see from the two builds that it doesn't push on non-master branches, and it doesn't push on pull requests (from forks).

asmeurer:
And it does work on master itself: https://travis-ci.org/gforsyth/travis_docs_builder/builds/146458586

PR #15, opened 2016-07-21
don't checkout to master after creating gh-pages branch

gforsyth:
the whole point is to BE ON THE GH-PAGES BRANCH


asmeurer:
I guess that works too. It probably shouldn't matter if the tracked files from master are laying around or not. The only files that matter are the built docs, which should be untracked on master.

gforsyth:
yeah, i scratched this stuff, i followed the logic again and realized this wasn't the actual issue.  about to open a separate PR just removing `--cached` from the remove

PR #19, opened 2016-07-21
Add a "tests" build to Travis (for now, just runs pyflakes)

asmeurer:


asmeurer:
Seems to work.  Now we just need real tests.

PR #23, opened 2016-07-22
Add basic argparse support

asmeurer:
A ton more options will need to be added. We should decide what the API should
look like.

Also, I am not married to argparse. If there is a argument parsing library
that you prefer, we can use that instead.

I can't even build a reasonable conda package out of version 1.0, since there's no way to run the script with a `--help` option.


asmeurer:
Any opposition to the command line so far? One thing I'm not sure about, should it do like it does now, where it detects if it is on Travis, and there is just one command `doctr` that does one thing locally and another thing on Travis (with flags to override that), or should there be subcommands like `doctr setup` and `doctr deploy` that you have to call explicitly? 

Beyond this I want to add flags to configure stuff, like the different directory locations and whatnot (but whatever can be detected automatically should be, by default).

gforsyth:
I have no opposition to the CLI stuff.  
As for whether or not to split it up... I think simpler is better (obviously) but I'm going back on forth on which is actually simpler for people to quickly grasp.  

I think I'm now leaning towards a split.  Just so it's more clear that "you run this command once" and then "travis runs this command all the time"

asmeurer:
Well I'm undecided as well. Here's a PR that splits it https://github.com/gforsyth/doctr/pull/28. I prefer the current `doctr` no args way, I'm just worried if it will be confusing to users.

PR #24, opened 2016-07-22
I'm curious if using Miniconda on Travis will be faster

asmeurer:


asmeurer:
Looks like it should be faster. Downloading miniconda and installing the packages seems to be faster than pip installing requests and cryptography, and as a bonus, Travis doesn't waste time installing Python 3.5.

asmeurer:
The Travis build statuses aren't being forwarded to GitHub for some reason. Maybe it will be fixed tomorrow.

gforsyth:
I had to update my travis profile to sync with the updated repo name.  Looks good!

Issue #6, opened 2016-07-17
Detect Sphinx build location automatically 


asmeurer:
Right now it just defaults to docs/_build/html, which is the default when you set up Sphinx, but a lot of people don't use the defaults.

asmeurer:
https://github.com/gforsyth/doctr/pull/65

PR #27, opened 2016-07-26
Detect docs

gforsyth:
Small addition to automatically detect where the sphinx directory is by searching recursively for a `conf.py` and backing out from there.  Maybe a little hacky.

Also removed a line advising users to create the `gh-pages` branch as this is handled automatically.  

In re: #6


doctr/__main__.py
```diff
@@ -63,9 +63,7 @@ def main():
 
 """.format(encrypted_variable=encrypted_variable.decode('utf-8'))
 
-        print("Put\n", travis_content, "in your .travis.yml.\n",
-            "Also make sure to create an empty gh-pages branch on GitHub, and "
-            "enable it at https://github.com/{repo}/settings".format(repo=repo), sep='')
```
asmeurer: They still need to enable it, though.

gforsyth: I just tested this -- I'm not sure when they changed it, but if you push up a `gh-pages` branch it is automatically enabled.

asmeurer: Oh great.

doctr/travis.py
```diff
@@ -186,3 +188,23 @@ def push_docs():
...
+    then backs out the name of the docs folder from there.
+
+    Output is printed for purposes of capture by bash on Travis
+    """
+    for conf_loc in glob.iglob('**/conf.py', recursive=True):
+        doc_loc = re.search(DOCS_REGEX, conf_loc)
+        if doc_loc:
+            doc_loc = doc_loc.group(1)
+            #print this to grab it in a BASH envvar
+            print(doc_loc)
```
asmeurer: It would be better to return the string. We can add a subcommand to print the output.

asmeurer:
Should change the keyword argument of https://github.com/gforsyth/doctr/pull/27/files#diff-d57991591acb44db37bc62c26e1748beR152 to default to `None` and call this function.

doctr/travis.py
asmeurer: I actually don't really see much value in getting this automatically just to build the docs. The individual project .travis.yml will know where its docs are. The real thing you need to find is the `_build` directory.

asmeurer: That may actually need a separate search, though, as I think Sphinx can be configured to build to `_build` or to `build` (run `sphinx-quickstart` and look at the second question it asks you).

gforsyth: I can grab the build directory pretty easily, since there's a copy of `conf.py` in there, too, so the same glob will find it, just need to split on that string differently.

asmeurer: I don't see a copy of conf.py in the build directory.

gforsyth: Yeah, actually, I don't either.  Must've been  a weird install that I was working on the other day.  I can glob for `build` then

asmeurer: So I guess it's either

```
index.rst
conf.py
_build
  |
  index.html
  ...
```
(like this repo)

or 

```
source
  |
  conf.py
  index.rst
build
  | 
  index.html
  ...
```

(like [this](https://github.com/sympy/sympy/tree/master/doc)).

Probably it could technically be anything, if you configure conf.py right, but those are the two default ways from `sphinx-quickstart`.

asmeurer:
You missed an import (see the tests)

gforsyth:
Hey @asmeurer -- this is ready for another look

doctr/travis.py
```diff
@@ -270,3 +274,27 @@ def push_docs():
...
+    """
+    for conf_loc in glob.iglob('**/conf.py', recursive=True):
+        doc_loc = re.search(DOCS_REGEX, conf_loc)
+        if doc_loc:
+            doc_loc = doc_loc.group(1)
+            try:
+                build_loc = glob.glob('{}/**build'.format(doc_loc),
+                                      recursive=True)[0]
+            except IndexError:
+                sys.exit('Cannot find build directory.  Is there a ``make html`` in your ``.travis.yml``?')
```
asmeurer: Should just raise a normal exception here (`RuntimeError` if you can't think of something better). We can catch exceptions in the command line to print nicer messages if you want.

asmeurer: This error message is confusing. It doesn't even look at .travis.yml

asmeurer:
I tried running the function in IPython and it didn't work.

doctr/travis.py
```diff
@@ -270,3 +274,27 @@ def push_docs():
         run(['git', 'push', '-q', 'doctr_remote', 'gh-pages'])
     else:
         print("The docs have not changed. Not updating")
+
+
+DOCS_REGEX = re.compile('([\w.]*)\/conf\.py')
+
+def find_build_dir():
```
asmeurer: Maybe `find_sphinx_build_dir`. We might eventually want to make doctr more generic than Sphinx.

doctr/travis.py
```diff
@@ -270,3 +274,27 @@ def push_docs():
...
+        if doc_loc:
+            doc_loc = doc_loc.group(1)
+            try:
+                build_loc = glob.glob('{}/**build'.format(doc_loc),
+                                      recursive=True)[0]
+            except IndexError:
+                sys.exit('Cannot find build directory.  Is there a ``make html`` in your ``.travis.yml``?')
+            build_loc = build_loc.split('/')
+
+            return build_loc
```
asmeurer: This returns a list. Shouldn't it return a path?

asmeurer:
Oh never mind, I hadn't built the docs yet.

asmeurer:
This doesn't work with the second configuration [here](https://github.com/gforsyth/doctr/pull/27#discussion_r72722352).

doctr/travis.py
gforsyth: I'll change it to a normal exception -- not sure about what to do for an error message.  It fails if there is no `build` directory, which usually means that sphinx hasn't run yet

gforsyth:
I need to rethink this a bit -- we don't need it for the first release and this is getting too convoluted.

asmeurer:
OK, but we at least need a flag to set it for the release, I'd say.

gforsyth:
Closing in favor of #65

PR #28, opened 2016-07-28
[Undecided] Split doctr into explicit 'doctr configure' and 'doctr deploy' subcommands

asmeurer:
It requires --force if travis (local) is run and it detects that it is (not)
on Travis.

I'm undecided if I actually want this behavior or not. The behavior before
this detects it automatically, allowing the user to switch the behavior if
so-desired. This way is more explicit.

In any case, it's worth noting that running the wrong command in the wrong
place almost never is what you want (except for us developers doing testing).
The local command requires user input. The Travis command requires the
encrypted GH_TOKEN variable, and also makes some assumptions that are always
true on Travis (like that it is on Linux, that it will be run from the git
repo directory, and that the git remote is https).


asmeurer:
FWIW, doing it this way makes the argparse code more complicated, although similar changes may eventually be needed anyway once we add flags to set various things on Travis.

gforsyth:
I'm also not sure.  Although I think it's a very good argument that having the command be less explicit makes it more user friendly.

asmeurer:
[Input from the masses](https://twitter.com/asmeurer/status/759137200890085376)

doctr/__main__.py
```diff
@@ -30,42 +30,67 @@
...
-    help="Run as if local (not on Travis). The default is to detect automatically.")
+
+    location = parser.add_subparsers(title='location', dest='location',
+        description="Location doctr is being run from")
+    travis_parser = location.add_parser('travis', help="Run as if on Travis.")
+    travis_parser.set_defaults(func=travis)
+    travis_parser.add_argument('--force', action='store_true', help="""Run the travis command even
+    if we do not appear to be on Travis.""")
+
+    local_parser = location.add_parser('local', help="Run as if local (not on Travis).")
```
asmeurer: Should probably call this `setup` or `configure`.

doctr/__main__.py
```diff
@@ -30,42 +30,67 @@
...
     parser.add_argument('-V', '--version', action='version', version='doctr ' + __version__)
-    location = parser.add_mutually_exclusive_group()
-    location.add_argument('--travis', action='store_true', default=None,
-    help="Run as if on Travis. The default is to detect automatically.")
-    location.add_argument('--local', action='store_true', default=None,
-    help="Run as if local (not on Travis). The default is to detect automatically.")
+
+    location = parser.add_subparsers(title='location', dest='location',
+        description="Location doctr is being run from")
+    travis_parser = location.add_parser('travis', help="Run as if on Travis.")
```
asmeurer: And call this one `deploy` or something like that.

asmeurer:
Twitter poll results:

screenshot 2016-08-01 19 52 04

I'm starting to think it's a good idea to separate it, as it will be cleaner once we start adding flags (which necessarily only make sense in one context). If anything, having the `doctr --help` split the flags by where they are used in this way is a strong argument for this.

asmeurer:
And apparently Travis detection can [fail](https://travis-ci.org/gforsyth/doctr/jobs/149056741).

asmeurer:
Any objections to this? I think I prefer it.

gforsyth:
Nope, works for me!

Issue #8, opened 2016-07-17
Use a more restrictive way of accessing GitHub


asmeurer:
Is it possible to register this thing on GitHub as an application, rather than as a personal access token?

asmeurer:
And more generally, if there is any way we can ask for fewer permissions than read/write access to all public repos, that would be great. Ideally we would just have access to push to the `gh-pages` branch of the one repo, but that sort of granularity only seems to be possible if we create a separate GitHub user. 

That's not at all ideal, but if it really is the only way, maybe there is some way we can do it. Or we can just stick with what we have now.

asmeurer:
Aha! So this is what matplotib uses: https://github.com/matplotlib/matplotlib/blob/master/.travis.yml

They use a deploy key (https://developer.github.com/guides/managing-deploy-keys/). This is an ssh key, which grants access to only a single repository. 

Here is the API: https://developer.github.com/v3/repos/keys/

asmeurer:
There is still some risk if the key is leaked, as they could push to other branches. This is limited if branch protection is enabled (specifically, required status checks). 

So we *definitely* need to implement this before we go public with this project. The current settings are way to open. A leaked token would grant someone write access to every single public repository for the person who created the token. 

We might keep the current way, because another option is to create a separate user for pushing the docs, and create a token for that user. The advantage to that is that GitHub will then let you grant access for that user to only push to `gh-pages`. Malicious parties could still do bad things if they got access to the token, but only to the gh-pages branch.

asmeurer:
The other downside: you need ssh stuff installed locally to generate a deploy key. I'm still trying to figure out how that works, actually.

asmeurer:
Argh! Travis supports this natively, but only for paid plans https://stackoverflow.com/questions/27444891/how-to-add-ssh-key-in-travis-ci

asmeurer:
This is how you generate a key https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/. You need to add `-f <key_location>` and `-P ''` to the ssh-keygen command to put the file in a given location, and to use an empty password (otherwise it will prompt for each). 

I wonder if it is possible to generate a key with a password, and encrypt that password into Travis. It seems like it would be doable, but probably wouldn't actually increase the security of anything. 

That generates a private a public key file. The private key file needs to be encrypted for Travis, and the public file needs to be added as a deploy key on GitHub. 

[...]

PR #30, opened 2016-07-31
Deploy key

asmeurer:
Work on fixing #8. Isn't finished yet.

### DIFF
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -29,7 +29,7 @@ from textwrap import dedent
 
 from .local import (generate_GitHub_token, encrypt_variable, encrypt_file,
     upload_GitHub_deploy_key, generate_ssh_key)
-from .travis import setup_GitHub_push, commit_docs, push_docs, get_repo
+from .travis import setup_deploy_key, setup_GitHub_push, commit_docs, push_docs, get_repo
 from . import __version__
 
 def main():
@@ -60,7 +60,9 @@ def main():
 
     if on_travis:
         repo = get_repo()
-        if setup_GitHub_push(repo):
+        if not args.token:
+            setup_deploy_key()
+        if setup_GitHub_push(repo, auth_type='token' if args.token else 'deploy_key'):
             commit_docs()
             push_docs()
     else:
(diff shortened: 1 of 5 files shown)


asmeurer:
I got the local stuff working. Now I just need to make it use the deploy key on Travis. 

I'm leaving the token stuff in, as it could still be useful if someone wants to create a custom GitHub user for uploading the docs.

asmeurer:
OK, now there is code to work with the deploy keys on Travis. It just needs testing.

asmeurer:
OK, @gforsyth, I can't add a deploy key to this repo. So you'll need to do it. Here's what to do:

- Review the code, see if you see any issues. If you don't merge it. Travis will fail in master.

- The reason is that I've revoked my token that this repo has been using thus far, and removed it from .travis.yml. 

- From master, run `python -m doctr` (the commands are still unsplit in this PR).  This will generate a deploy key and upload it to GitHub. 

- Follow the instructions from the command (and let me know if they could be improved). You need to commit a file, and add an encrypted environment variable to .travis.yml.

- Create a new PR with those changes and merge it. 

If everything works right, the docs should deploy (you'll be able to tell because all the new functions I added here will be in the API docs). If not, we'll have to iterate. Unfortunately, only you can modify the deploy key on GitHub. The code by design destroys the private key once it's uploaded to GitHub, and GitHub also prevents you from seeing it, so if something goes wrong, you'll have to regenerate everything from scratch.

doctr/__main__.py
```diff
@@ -47,25 +60,57 @@ def main():
...
+                {ssh_key}
+
+                Be sure to allow write access for the key.
+                """.format(ssh_key=ssh_key, deploy_keys_url=deploy_keys_url)))
+
+            # TODO: Should we just delete the public key?
+
+            print(dedent("""Commit the file github_deploy_key.enc. The file
+            github_deploy_key.pub contains the public deploy key for GitHub.
+            It does not need to be committed."""))
```
asmeurer: Should we just delete the public key? It's either uploaded to GitHub or printed to the screen. It's harmless if committed, but unimportant.

PR #32, opened 2016-08-01
add dash in travis.yml and also in explanation text

gforsyth:
it was missing in the helper text and so travis defaults to a token
based deploy (and then fails)


asmeurer:
It worked before like that. Did this actually fix the problem?

gforsyth:
Nope.  I think I just found it though.  When `DOCTR_DEPLOY_ENCRYPTION_KEY` exists then `token = ''` which errors out when it hits a `decode`.  Just pushed up a fix

asmeurer:
Ah damn. The "this should be bytes" vs. "this should be str" trips up the best of us. If only Python had a type checker.

gforsyth:
New project:  A Travis CI based Python type checker

PR #33, opened 2016-08-01
encode empty string token

gforsyth:
explicitly encodes empty string for `token` if we're using a deploy key


doctr/travis.py
```diff
@@ -91,7 +91,7 @@ def run(args):
     Automatically hides the secret GitHub token from the output.
     """
     if "DOCTR_DEPLOY_ENCRYPTION_KEY" in os.environ:
-        token = ''
+        token = ''.encode('utf-8')
```
asmeurer: You can just write `b''`.

gforsyth: incoming in the next fix

PR #35, opened 2016-08-01
fix typo and simplify empty string encoding

gforsyth:


doctr/travis.py
```diff
@@ -50,7 +50,7 @@ def setup_deploy_key():
         raise RuntimeError("DOCTR_DEPLOY_ENCRYPTION_KEY environment variable is not set")
 
     key = key.encode('utf-8')
-    decrypt_file('githib_deploy_key.enc', key)
+    decrypt_file('github_deploy_key.enc', key)
```
asmeurer: My next change will be to globally factor this out into a variable (it needs to be configurable anyway).

PR #37, opened 2016-08-02
Create .ssh on Travis if it doesn't exist

asmeurer:
Hopefully that's what the problem is with the latest failing builds.


asmeurer:
I guess that wasn't the problem. Still, a good idea to do this.

PR #43, opened 2016-08-02
Try ssh fix

asmeurer:
(just want to follow the changes to this branch)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ install:
   - source activate test-environment
 
 script:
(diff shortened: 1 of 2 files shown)


gforsyth:
Ok -- I _think_ this is working now.  We'll see what happens.

gforsyth:
Sigh... nope.  The easy workaround here is for people to put ``eval `ssh-agent -s``` in their `travis.yml` but I think I'm close to getting it down automatically.  But I've thought that for the last hour or so

PR #45, opened 2016-08-02
move encrypted key thing to inside matrix

gforsyth:


gforsyth:
well, that doesn't bode well...

asmeurer:
You can use `travis lint` from the Travis gem to check for basic errors in the .travis.yml.

PR #46, opened 2016-08-05
Make the key path configurable

asmeurer:
I also took the liberty of making the `doctr configure` instructions a little more user friendly.


doctr/__main__.py
```diff
@@ -53,6 +53,8 @@ def main():
     deploy_parser.add_argument('--token', action='store_true', default=False,
         help="""Push to GitHub using a personal access token. Use this if you
         used 'doctr configure --token'.""")
+    deploy_parser.add_argument('--key-path', default='github_deploy_key.enc',
+        help="""Path of the encrypted GitHub deploy key. The default is '%(default)s'.""")
```
gforsyth: Oh wow.  I didn't know you could do the `%(default)s` thing.  That's awesome.

PR #49, opened 2016-08-05
also install doctr

gforsyth:
For anyone not us, they need to install doctr in the `.travis.yml` file or things will go bonk.

Also, our example `.travis.yml` is doing things with miniconda -- do we want to mention this in the README or just stick with `pip`?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,7 +40,7 @@ secure key to add. You should also have something like
        secure: "<your secure key from doctr here>"
 
    script:
-     - pip install requests cryptography sphinx doctr
+     - pip install sphinx doctr
      - cd docs
      - make html
      - cd ..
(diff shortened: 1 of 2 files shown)


asmeurer:
pip is fine unless you can't build your docs with Python 3.5. Then conda is possibly easier (I'm not sure if it's possible to install multiple Pythons in the same build in Travis).

README.rst
```diff
@@ -40,7 +40,7 @@ secure key to add. You should also have something like
        secure: "<your secure key from doctr here>"
 
    script:
-     - pip install requests cryptography sphinx
+     - pip install requests cryptography sphinx doctr
```
asmeurer: You don't need to install requests or cryptography. Those are dependencies of doctr and will be installed automatically.

asmeurer: Should also add 

    - pip install doctr

to the instructions in `__main__.py`.

PR #65, opened 2016-08-12
Find docs

gforsyth:
Second, simpler shot at detecting sphinx build location automatically, superceding #27

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -229,10 +229,13 @@ def create_gh_pages():
 
 def find_sphinx_build_dir():
     """
-    Find name of build folder within sphinx docs directory.
+    Find name of docs folder and build subfolder within sphinx docs directory.
 
     Locate the sphinx docs directory by looking for a ``conf.py`` file.
     Then look for the ``build`` folder, by convention either ``build`` or ``_build``
+
+    This is called by :func:`commit_docs` if keyword arg ``built_docs`` is not
+    specified on the command line.
     """
     docs_folder = glob.glob('**/conf.py')[0].split('/', 1)[0]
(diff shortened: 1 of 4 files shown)


doctr/travis.py
```diff
@@ -226,20 +227,37 @@ def create_gh_pages():
...
+    """
+    Find name of docs folder and build subfolder within sphinx docs directory.
+
+    Locate the sphinx docs directory by looking for a ``conf.py`` file.
+    Then look for the ``build`` folder, by convention either ``build`` or ``_build``
+
+    This is called by :func:`commit_docs` if keyword arg ``built_docs`` is not
+    specified on the command line.
+    """
+    docs_folder = glob.glob('**/conf.py')[0].split('/', 1)[0]
```
asmeurer: This needs better error handling when the glob fails.

asmeurer: You need `recursive=True` for the `**` glob to work.

gforsyth: Are you sure about the `recursive` kwarg?  It works locally for me without it.  I guess I could just make it a `*/conf.py` since that file should only be one level deep

asmeurer:
OK, fixed those issue. I tested it, and it works with this repo and with SymPy (which uses the alternate conf.py setup).

gforsyth:
Sweet!

doctr/travis.py
asmeurer: It wasn't working for me without it. That's also what it says at https://docs.python.org/3/library/glob.html

asmeurer:
This failed https://travis-ci.org/gforsyth/doctr/jobs/151897733

PR #66, opened 2016-08-12
Add a simple test for find_sphinx_build_dir()

asmeurer:
Also, enable tests in Travis.

Feel free to push other tests, or otherwise to merge.


asmeurer:
Just realised, this test will fail if you install doctr. Should probably move it out of the library.

asmeurer:
Oh derp, this test can't pass unless the docs have been built. Maybe I should just scrap this.

PR #71, opened 2016-08-16
change means of locating sphinx build directory

### LINKED ISSUE #69: Master failing (find docs)
See https://travis-ci.org/gforsyth/doctr/jobs/151897733

gforsyth:
This should hopefully resolve #69

If `doctr deploy` is run for the first time on a repository, the old
method works fine -- it locates `conf.py` and then goes from there.
However, if it isn't the first time, then an orphan `gh-pages` branch
has already been created and the `conf.py` file has been deleted in this
branch, so the location glob will fail.

This instead looks for `*build/html` recursively, since we are assuming
the build directory to be named either `_build` or `build`

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -229,24 +229,17 @@ def create_gh_pages():
 
 def find_sphinx_build_dir():
     """
-    Find name of docs folder and build subfolder within sphinx docs directory.
-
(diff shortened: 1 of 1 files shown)


asmeurer:
Looks good. Let's try it.

asmeurer:
It worked.

Issue #10, opened 2016-07-18
Default should be the root, not /docs


asmeurer:
The default should be the root, not /docs, since this is supposed to be for simple projects where the docs are the homepage, or, alternately, a `docs` repo in an org. 

It's harder, though, because you have to do an rsync to update stuff. Maybe there's a simpler way to do it.

gforsyth:
@asmeurer -- I could well be wrong, but I thought that if the target is docs then sphinx creates an index file that points to it, no?  Maybe I configured that separately somewhere...

asmeurer:
I've done that sort of thing manually before with a redirect page, like

```html
<html>
    <head>
        <meta http-equiv="refresh" content="0;URL=docs/index.html" />
    </head>
    <body>
    </body>
</html>
```

(apparently doing redirects that way is frowned upon, but it's the only way to do it with static pages). I would leave that to the users to do that if they want, although we can document that this is a thing in our docs.

asmeurer:
I guess instead of rsync you could clear every file tracked by git. You'll also probably need a way to specify files to not clear (with `.nojekyll` being left alone by default).

asmeurer:
Or when it commits the docs, it could also write a file like `.doctr-log` to the root with a list of every file that was committed, so that when it commits files again it can use that file to know which files to clear.

asmeurer:
My goal is to have this and #25 for a 1.2 release (and we already have #12). 

After that, pretty much everything that I personally would like is implemented. #29 would be nice, but I'm a little unclear on what exactly to do there. Writing some tests (#5) would also be nice, but it seems like just about everything in the library is difficult to test, because it relies on some external state like an encrypted environment variable or an internet connection. You can do mock testing, but it's difficult to know, for instance, if something will actually work on Travis unless you actually run it on Travis.

PR #73, opened 2016-08-17
Allow docs to be deployed to .

asmeurer:
This implements syncing with a log file. 

Fixes https://github.com/gforsyth/doctr/issues/10 (although I haven't changed the default)


asmeurer:
I think this works, but since it's actually testable, I'd like to write some tests for it before merging.

asmeurer:
OK, I added some tests.

Issue #67, opened 2016-08-12
Move to a github organization?


gforsyth:
`doctravis` is available.  
other options:
`drdoctr`
`rtcod`
`youwillfeelasmallpinch`

asmeurer:
`rtcod` is just evil

asmeurer:
I like `drdoctr`, although I would feel strongly compelled to create a subrepository called `givemethenews`.

gforsyth:
Ok, sounds good to me.

gforsyth:
organization created and I sent you an invite.  I won't transfer the repo just yet -- maybe we want to wait until we have `github.io` support working?

asmeurer:
Maybe we should add support to deploy to multiple destinations, so that we can test all three. Right now, you can only deploy to one place because there can only be one environment variable with the decryption key.

PR #75, opened 2016-08-18
Update changelog for upcoming 1.2 release

asmeurer:
I think it should be 1.2 instead of 1.1.2, as there have been several major
changes.

I have also added the org moving to the changelog
(https://github.com/gforsyth/doctr/issues/67), since I'm assuming that will
happen soon (a PR will be needed; several places in the code reference
"gforsyth", and we'll probably need a new encrypted deploy key).


gforsyth:
I'm going to merge in all the current PRs, then transfer the repository to the organization and then put in a PR with all the naming updates and the new deploy key.

Issue #74, opened 2016-08-18
Add some recipes to the docs


asmeurer:
For example, deploying to a directory based on the branch name.

PR #76, opened 2016-08-18
Add some recipes to the docs

asmeurer:
Fixes https://github.com/gforsyth/doctr/issues/74.

If you have any suggestions to add, feel free.


asmeurer:
A repo.github.io style gh-pages recipe would be nice too (if just to document that it "just works").

PR #78, opened 2016-08-18
Migration

gforsyth:
This change the name of the organization everywhere in the code and in the docs.  Also adding the new deploy key for the organization


asmeurer:
Can you add a redirect to https://gforsyth.github.io/doctr/docs/. It will break the GitHub redirect when you create github.com/gforsyth/doctr, but https://gforsyth.github.io/doctr/docs/ is the link I've been sharing, so it would be better.

gforsyth:
Redirect added to both `doctr` and `doctr/docs` and I also added a note to the README about the new repo location

asmeurer:
Great, thanks.

PR #83, opened 2016-08-18
When deploying to ., only commit the files that from the source docs

asmeurer:
Fixes https://github.com/drdoctr/doctr/issues/82

I still want to add some tests.


asmeurer:
Ready for review. Before merging, we should clear the `gh-pages` branch, or the build will keep failing.

gforsyth:
makes sense to just delete the branch, yeah?

gforsyth:
Ok -- this looks good to me.  I'm going to delete `gh-pages` and then merge this

PR #89, opened 2016-08-30
1.2

asmeurer:
No movement has happened on the github.io pages stuff, so I'm just going to do a release with what's there.


asmeurer:
[Travis is down](https://www.traviscistatus.com/#). It says tests passed but I want to load the page to make sure. So once it starts working I'll take the release.

PR #94, opened 2016-08-30
We aren't using the Travis Python, maybe it's faster to not define a language

asmeurer:


asmeurer:
No, because then it uses Ruby.

PR #100, opened 2016-08-31
Add link to GH docs on branch protection

willingc:


asmeurer:
Argh. It looks like I somehow broke Travis tests on PRs from forks. Let me see if I can fix it.

asmeurer:
@willingc can you click the "update with master" button?

asmeurer:
Thanks for the patch. And I wouldn't have noticed that really bad bug as quickly if it weren't for this (though I really should have noticed it last night).

willingc:
Glad to be of service ;-) @carreau mentioned the project to me yesterday. I like it 👍

asmeurer:
Thanks. Would be happy to hear any feedback you have if you use it.

PR #108, opened 2016-09-01
Fix repo check

### LINKED ISSUE #107: Repo check fails for forks
We should probably be using [this api](https://developer.github.com/v3/repos/) to check if a repo exists.

asmeurer:
The search doesn't find forks. This removes the distinction between a wrong
user and a wrong repo, but I think that's fine.

Fixes #107.

### DIFF
--- a/doctr/local.py
+++ b/doctr/local.py
@@ -187,12 +187,10 @@ def check_repo_exists(deploy_repo):
         raise RuntimeError('"{deploy_repo}" should be in the form username/repo'.format(deploy_repo=deploy_repo))
 
     user, repo = deploy_repo.split('/')
-    search = 'https://api.github.com/search/repositories?q={repo}+user:{user}'
-    r = requests.get(search.format(user=user, repo=repo))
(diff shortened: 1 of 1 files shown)


asmeurer:
Don't know why that's failing intermittently.

asmeurer:
But I think it doesn't work on forks anyway (https://github.com/drdoctr/doctr/issues/110). So maybe we can keep the existing check but change the error message. I'm not sure what to do here.

asmeurer:
And it passed this time. Still, and intermittent failure like that is no good. But maybe it was an issue on GitHub's side?

gforsyth:
Hey @asmeurer , I'm back in the world.  I just triggered a rebuild -- I think if that works then we can chalk it up as a (resolved) github issue?

asmeurer:
I guess it's a GitHub issue. If the API were wrong it wouldn't work at all.

gforsyth:
Ok, merging this

PR #120, opened 2016-09-15
swap description and long description

### LINKED ISSUE #119: "summary" and "description" fields on pypi should be switched around
A package's summary should be a short one-liner which gives people an idea what the package is about. However, when you do `pip search doctr`, you get a unreadable wall of text, with the formatting broken and the newlines removed. The wall of text should go into the `description` field, which preserves the formatting.

gforsyth:
Resolves #119

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,19 +11,12 @@ setup(
     name='doctr',
     version=versioneer.get_version(),
     cmdclass=versioneer.get_cmdclass(),
-    description=open("README.rst").read(),
     author='Aaron Meurer and Gil Forsyth',
     author_email='asmeurer@gmail.com',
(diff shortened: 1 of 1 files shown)


asmeurer:
It's that damn test again. Maybe we should revert that PR.

gforsyth:
Ok, I restarted the test and it passed this time.  Maybe just a weird github API thing again?  Or are we hitting some limits for unauthorized clients?  I'm ok reverting the tests if they continue to cause trouble.

asmeurer:
Looks better now https://pypi.python.org/pypi/doctr

gforsyth:
That looks *much* better.

asmeurer:
:)

PR #121, opened 2016-09-18
Add support for private GH repos

dan-blanchard:
This PR makes it so you can use doctr with private GitHub repos, which was previously impossible.

I had to move a few things around for this to work, like requiring you to login to GH when you first start things up instead of in `GitHub_post`, because `check_repo_exists` needs to be an authenticated check to look for private repos.

This also creates a temporary GH token to login to Travis (because Travis doesn't support anything but GH tokens for login), which is deleted immediately after we use it.

### DIFF
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -89,7 +89,9 @@ options available.
         public repositories for the user. This option is not recommended
         unless you are using a separate GitHub user for deploying.""")
     configure_parser.add_argument("--no-upload-key", action="store_false", default=True,
-        dest="upload_key", help="""Don't automatically upload the deploy key to GitHub.""")
+        dest="upload_key", help="""Don't automatically upload the deploy key to GitHub. If you select this
+        option, you will not be prompted for your GitHub credentials, so this option is not compatible with
+        private repositories.""")
     configure_parser.add_argument('--key-path', default='github_deploy_key',
         help="""Path to save the encrypted GitHub deploy key. The default is %(default)r.
     The .enc extension is added to the file automatically.""")
@@ -167,7 +169,10 @@ def configure(args, parser):
         parser.error("doctr appears to be running on Travis. Use "
             "doctr configure --force to run anyway.")
 
-    login_kwargs = GitHub_login()
+    if args.upload_key:
+        login_kwargs = GitHub_login()
+    else:
+        login_kwargs = {'auth': None, 'headers': None}
 
     build_repo = input("What repo do you want to build the docs for (org/reponame, like 'drdoctr/doctr')? ")
     is_private = check_repo_exists(build_repo, **login_kwargs)
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -186,7 +186,7 @@ def configure(args, parser):
     if args.token:
         token = generate_GitHub_token(**login_kwargs)['token']
         encrypted_variable = encrypt_variable("GH_TOKEN={token}".format(token=token).encode('utf-8'),
-            build_repo=build_repo, is_private=is_private, gh_token=token, **login_kwargs)
+            build_repo=build_repo, is_private=is_private, **login_kwargs)
         print(dedent("""
         A personal access token for doctr has been created.
 
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -30,7 +30,7 @@ import subprocess
 from textwrap import dedent
 
 from .local import (generate_GitHub_token, encrypt_variable, encrypt_file,
-    upload_GitHub_deploy_key, generate_ssh_key, check_repo_exists)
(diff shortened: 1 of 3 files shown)


dan-blanchard:
It might be worth adding a warning about how all GH pages are public, so even if you're repo is private, everyone can still see its pages, but I'm not sure where we'd want that.

doctr/__main__.py
```diff
@@ -89,8 +89,7 @@ def get_parser():
         public repositories for the user. This option is not recommended
         unless you are using a separate GitHub user for deploying.""")
     configure_parser.add_argument("--no-upload-key", action="store_false", default=True,
-        dest="upload_key", help="""Don't automatically upload the deploy key to GitHub. If you select this
-        option, you will not be prompted for your GitHub credentials. """)
+        dest="upload_key", help="""Don't automatically upload the deploy key to GitHub.""")
```
asmeurer: Can we keep this? There's a valid [use-case](https://drdoctr.github.io/doctr/recipes.html#setting-up-doctr-for-a-repo-you-don-t-have-admin-access-to) for this.

asmeurer:
So this works with both the token and deploy key?

doctr/__main__.py
dan-blanchard: All that's changed is that you'll always be prompted for a password, because `check_exists_repo` needs to be authenticated for private repos to work. The use-case you linked doesn't to be broken by the user having to login.

asmeurer: Maybe we can ask if the repo is private if it doesn't seem to exist, and then prompt?

dan-blanchard:
I didn't test it with token at first, but I just did and it looks good after removing a little bit of cruft from an earlier approach I was taking to change `encrypt_variable.

doctr/__main__.py
dan-blanchard: Hmm... That's reasonable, but I think we should probably only do that in the `--no-upload-key` case, since we'll need the login info saved to pass to `GitHub_post` and `encrypt_variable` and a bunch of other functions otherwise.

dan-blanchard: How about we just don't allowed `--no-upload-key` with private repos? Trying to prompt for password if it's not found is going to lead to a mess, because we'll need to get the `login_kwargs` out of `check_repo_exists` then so we have them to pass to other functions down the line.  Seems to me that `--no-upload-key` exists for people who want to add doctr docs to a public repo that they don't have access to, whereas if they have access to a private repo, they can probably upload keys no problem.

asmeurer: That's fine.

dan-blanchard: Okay, I did that in c54ef45.

asmeurer:
Should we make a distinction between private on GitHub and using travis-ci.com? I think it's possible to use travis-ci.com for public repos. 

Also, if you're using travis-ci.com, there's a much better way to manage the deploy key, which is to upload it directly to Travis (https://docs.travis-ci.com/user/private-dependencies/, https://docs.travis-ci.com/api#settings:-ssh-key). That avoids the shenanigans we have to do to decrypt it, and doesn't require editing .travis.yml.

dan-blanchard:
> I think it's possible to use travis-ci.com for public repos.

I don't believe so. Everything on their [plans page](https://travis-ci.com/plans) points to travis-ci.org for public repos.  Even if it is, private on GitHub entails private on travis-ci.com, so I don't think we need to separate the two.

> Also, if you're using travis-ci.com, there's a much better way to manage the deploy key, which is to upload it directly to Travis (https://docs.travis-ci.com/user/private-dependencies/, https://docs.travis-ci.com/api#settings:-ssh-key). That avoids the shenanigans we have to do to decrypt it, and doesn't require editing .travis.yml.

That's true. I was just trying to make the smallest set of changes to get things working.

asmeurer:
> I don't believe so. Everything on their plans page points to travis-ci.org for public repos. Even if it is, private on GitHub entails private on travis-ci.com, so I don't think we need to separate the two.

I guess I was mistaken then. 

> That's true. I was just trying to make the smallest set of changes to get things working.

We can postpone it to a further PR if you want. I wish they would implement this feature for travis-ci.org. Then we could scrap what we are currently doing and do it the simpler (and likely more secure) way.

dan-blanchard:
Yeah, I'd rather postpone it for now.

asmeurer:
OK. I like that there are no changes to travis.py here. Unfortunately, local.py is the least tested (automatically). I'll run through the steps (for public repos, I don't have any private repos to test against) to make sure nothing is obviously broken, then merge.

asmeurer:
@dan-blanchard I've released doctr 1.3.3 with this change, and it should be on conda-forge shortly.

asmeurer:
And I opened https://github.com/drdoctr/doctr/issues/124 for the direct key upload thing. I don't have any private repos to test against, so if you want it you'll have to implement it (or someone else will).

PR #123, opened 2016-09-20
check that build repo exists on travis

### LINKED ISSUE #114: Check for travis repo before generating keys.
Otherwise it will just fail and you'll have to regenerate. 

It's a small optimisation though, that will mostly affect users that make typos.

Note that github is not case sensitive for usernames, travis is so for example, I regularly get into trouble when I write my username lowercase and it involves travis.

gforsyth:
resolves #114


gforsyth:
ok, I think it's time to remove those tests

asmeurer:
Can we add tests for it.

asmeurer:
Should I hold off [the release](https://github.com/drdoctr/doctr/pull/122) for this (I was going to do it today)?

gforsyth:
I can add tests similar to the problematic ones for github...?  Might be that they won't barf as often.  I'm not quite sure how else to do it

asmeurer:
We should revert the PR that broke the tests. IIRC the change is not that important.

gforsyth:
hrr, I just realized this may have ramifications for the just merged private repository stuff.  I would say go ahead with the release and we can always do a microrelease in a few days

asmeurer:
And maybe we should contact GitHub about it. We are either using their API wrong, or they have a bug.

gforsyth:
Ok, I'll raise an issue for that

gforsyth:
Which PR was it when the tests started breaking?

gforsyth:
Hmm... I wonder if it's a rate limiting thing?  if the PR build and the PUSH build arrive at the tests at the same time and then each hit the unauthed API 3 times in quick succession?  Because restarting the jobs individually always seems to work

asmeurer:
This one https://github.com/drdoctr/doctr/pull/108. I made the change because the old way doesn't work with forks (and seemed hacky anyway). But I later found out that Travis won't deploy to forks anyway (it disables encrypted environment variables on forks, not just builds from a PR on a fork but any build on a repo that's a fork). So the change isn't really necessary.

asmeurer:
Maybe we should add a small retry loop with exponential backoff, like

```
RETRIES = 5
wait = 1
for i in range(RETRIES):
    sleep(wait)
    run_check()
    wait *= 2
```

gforsyth:
This is ready for another look now that the github rate limiting thing is resolved.

PR #126, opened 2016-09-23
Run everything except for the push on PRs and non-master

### LINKED ISSUE #125: Run almost all doctr steps on PR
The only thing doctr shouldn't do is upload to gh-pages.

If, as part of your build, you modify your source tree, for example, doctr will refuse to do checkouts for it.  The same happens for incorrect file permissions (say arising from running sphinx inside a docker container).

Right now doctr just skips when doing PR builds it seems.  This means that these kinds of failures are not detected at the PR level and happen only when TravisCI is doing the push build.

For example  
- https://travis-ci.org/MaxPoint/spylon/builds/160663846
- https://travis-ci.org/MaxPoint/spylon/builds/160656976

By running everything but the actual git push that doctr does we can minimize this friction.

asmeurer:
Fixes #125.

This is a test to see if it works. If it does, we may want to restructure this
a bit to be a little cleaner (in particular, print the message about why it
isn't pushing at the end).

### DIFF
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -125,31 +125,34 @@ def deploy(args, parser):
 
     current_commit = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('utf-8').strip()
     try:
-        if setup_GitHub_push(deploy_repo, auth_type='token' if args.token else
+        can_push = setup_GitHub_push(deploy_repo, auth_type='token' if args.token else
                              'deploy_key', full_key_path=args.key_path,
-                             require_master=args.require_master):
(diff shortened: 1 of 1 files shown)


asmeurer:
So it's not that easy. Some configuration that happens in setup_GitHub_push needs to happen (i.e., the git config stuff).

Carreau:
One more point to get that, you want doctr to run also on non-canonical repo and just don't push. 
I assume the documentation repo is public, so clonning/fetching it is likely possible.

asmeurer:
This was finished at https://github.com/drdoctr/doctr/pull/132

PR #127, opened 2016-10-08
Use latest version of sphinxcontrib.autoprogram

asmeurer:
Our workaround is no longer necessary. The newer version also shows the usage
nicer, and doesn't show suppressed flags.


asmeurer:
Before: https://drdoctr.github.io/doctr/commandline.html
After: https://drdoctr.github.io/doctr/docs-autoprogram/commandline.html

gforsyth:
Looks good!

PR #130, opened 2016-10-20
Add a note to the README that doctr does not require Sphinx

asmeurer:
Fixes https://github.com/drdoctr/doctr/issues/129


asmeurer:
Oh it's that goddamn test again.

gforsyth:
Ugh.  Yeah -- I'm working on allowing doctr to deploy to other branches and I think when I pushed up my work we hit whatever rate limit github is imposing.

PR #132, opened 2016-11-02
No push behavior

Carreau:
based on @asmeurer #126

### DIFF
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -78,6 +78,10 @@ options available.
         conjunction with the --command flag, for instance, if the command syncs
         the files for you. Any files you wish to commit should be added to the
         index.""")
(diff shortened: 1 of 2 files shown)


doctr/__main__.py
```diff
@@ -148,7 +152,7 @@ def deploy(args, parser):
 
         changes = commit_docs(added=added, removed=removed)
         if changes:
-            if can_push:
+            if can_push and not args.no_push:
```
asmeurer: So even if the option is passed, do we still need to run `setup_GitHub_push` (we can't write `can_push = args.push and setup_GitHub_push(...)` above)?

asmeurer:
Syntax error :)

Carreau:
Great ! Thanks ! Sorry I'l be low bandwith this week.

PR #133, opened 2016-11-07
Authentication for GH tests

gforsyth:
I created a personal access token for GitHub that has no permissions and added it to the doctr travis org so it can be used to authenticate requests to GH for the tests.  This changes the rate limiting from 60/hour to 5000/hour and so should eliminate the annoying test failures we've had to deal with.

### DIFF
--- a/docs/tests.rst
+++ b/docs/tests.rst
@@ -20,4 +20,4 @@ You can add the updated token to Travis on the `doctr Travis Settings Page
 
 Paste the token string into the ``Value`` field and ``TESTING_TOKEN`` in the
 ``Name`` field (unless you have changed this value in
-``doctr/tests/test_local.py``
+``doctr/tests/test_local.py``).
--- a/docs/tests.rst
+++ b/docs/tests.rst
@@ -1,0 +1,23 @@
+GitHub Test Authentication
+--------------------------
+
+Many of the ``doctr`` tests in ``doctr/tests/test_local.py`` contact GitHub to
+check that invalid repo names raise errors, etc. However, the GitHub API has a
+hard limit of 60 requests / hour for unauthenticated requests and it is very
+easy to run over this when pushing many changes.
+
+In order to avoid this limit, there is a GitHub Personal Access Token stored in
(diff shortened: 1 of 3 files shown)


asmeurer:
Can you add a comment or something on how to recreate the token should that be necessary in the future.

gforsyth:
Will do

doctr/tests/test_local.py
```diff
@@ -1,21 +1,27 @@
+import os
+
 from ..local import check_repo_exists
 
 from pytest import raises
 
+TEST_TOKEN = os.environ.get('TESTING_TOKEN')
+HEADERS = {'Authorization': 'token {}'.format(TEST_TOKEN)}
```
asmeurer: Does this still work if the token is empty?

gforsyth: No, I just added a fix for that.

gforsyth:
Added a docs page on regenerating the token.  It should probably be part of some larger `contributing` page but we can roll that out later

docs/tests.rst
```diff
@@ -0,0 +1,23 @@
...
+<https://github.com/settings/tokens>`_ and create a Personal Access Token. Make
+sure that all of the checkboxes are unchecked (this token should only have
+privileges to check in with the GitHub API).
+
+You can add the updated token to Travis on the `doctr Travis Settings Page
+<https://travis-ci.org/drdoctr/doctr/settings>`_.
+
+Paste the token string into the ``Value`` field and ``TESTING_TOKEN`` in the
+``Name`` field (unless you have changed this value in
+``doctr/tests/test_local.py``
```
asmeurer: Missing closing paren and period.

asmeurer:
The token is not encrypted on Travis. No concerns about that?

gforsyth:
I believe it is encrypted in the same fashion as the deploy key as per https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings

asmeurer:
Ah OK. I forgot that encrypted keys were possible from the web interface. This looks fine.

PR #134, opened 2016-11-08
Add more information to the automated commit message

asmeurer:
It now includes the repo name, branch, and commit.

We could add other things, but that would require passing in information to
commit_docs().

For reference, the list of all environment variables in Travis is at
https://docs.travis-ci.com/user/environment-variables/#Convenience-Variables.

Fixes https://github.com/drdoctr/doctr/issues/105 (although I might want to include more info than this).

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -338,13 +338,15 @@ def commit_docs(*, added, removed):
 Update docs after building Travis build {TRAVIS_BUILD_NUMBER} of
 {TRAVIS_REPO_SLUG}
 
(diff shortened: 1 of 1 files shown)


asmeurer:
[Example commit](https://github.com/drdoctr/doctr/commit/690d24e72f761b2ea7d7d6d8b563c285db25a6e7). 

Any other information we should include?

gforsyth:
Nothing comes to mind immediately.  We can always add more later, but this is a big improvement in any case.

PR #135, opened 2016-11-08
add `init` as alias for `configure`

### LINKED ISSUE #113: configure -> init ?
Hey guys !

Finally get some time to play along with doctr. 

Is there a reason why you choose `configure` over `init`, most of tool these days une `init`: the travis gem, git, mercurial, flit, npm, Cargo (rust) ...

I think that having init (at least as an alias) would be nice.

gforsyth:
resolves #113

### DIFF
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -81,9 +81,10 @@ options available.
     deploy_parser.add_argument('--no-push', dest='push', action='store_false',
         default=True, help="Run all the steps except the last push step."
         "Useful for debugging")
-        
 
-    configure_parser = subcommand.add_parser('configure', help="Configure doctr. This command should be run locally (not on Travis).")
+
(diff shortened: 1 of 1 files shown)


asmeurer:
Looks like a bug with sphinxcontrib-autoprogram. Might want to report that [here](https://bitbucket.org/birkenfeld/sphinx-contrib).

gforsyth:
Reported here: https://bitbucket.org/birkenfeld/sphinx-contrib/issues/174/autoprogram-duplicate-id-warning-raised

This doesn't error out in autoprogram 0.1.2 if we feel like downgrading until this issue is fixed.

asmeurer:
If we do that we'd have to revert to the other workaround we were using. Let's wait to see if the bug you reported gets fixed quickly.

PR #137, opened 2016-11-09
Start using .travis.yml as a configure doctr.

Carreau:
My guess is it will be easier with a subparser. 

Just playing with it for now, see what's needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,3 +44,4 @@ doctr:
   key-path: deploy_key.enc
   require-master: true
   sync: False
+  lubalubadubdub: False
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
   - conda config --add channels conda-forge # For sphinxcontrib.autoprogram
   - conda update -q conda
   - conda info -a
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION requests cryptography sphinx pyflakes sphinxcontrib-autoprogram pytest sphinx-issues
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION requests cryptography sphinx pyflakes sphinxcontrib-autoprogram pytest sphinx-issues pyyaml
   - source activate test-environment
 
 script:
@@ -31,12 +31,17 @@ script:
       cd docs;
       make html;
       cd ..;
-      python -m doctr deploy --key-path deploy_key.enc .;
-      python -m doctr deploy --key-path deploy_key.enc --gh-pages-docs docs;
-      python -m doctr deploy --no-require-master  --built-docs docs/_build/html --key-path deploy_key.enc "docs-$TRAVIS_BRANCH";
-      python -m doctr deploy --no-require-master --key-path deploy_key.enc --no-sync --command "echo test" docs;
+      python -m doctr deploy --sync .;
+      python -m doctr deploy --sync --gh-pages-docs docs;
+      python -m doctr deploy --sync --no-require-master  --built-docs docs/_build/html "docs-$TRAVIS_BRANCH";
+      python -m doctr deploy --no-require-master  --command "echo test" docs;
     fi
   - if [[ "${TESTS}" == "true" ]]; then
       pyflakes doctr;
       py.test doctr;
     fi
+doctr:
+  key-path: deploy_key.enc
+  require-master: true
+  gh-pages-docs: 'docs'
+  sync: False
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,5 +43,5 @@ script:
 doctr:
   key-path: deploy_key.enc
   require-master: true
-  gh-pages-docs: 'docs'
+  deploy-directory: 'docs'
   sync: False
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,5 +43,4 @@ script:
 doctr:
   key-path: deploy_key.enc
   require-master: true
-  deploy-directory: 'docs'
   sync: False
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,7 +7,13 @@ Current
 - The ``--gh-pages-docs`` flag of ``doctr deploy`` has been deprecated.
   Specify the deploy directory like ``doctr deploy .`` or ``doctr deploy docs``.
   There is also no longer a default deploy directory. (:issue:`128`)
-
+- ``setup_GitHub_push`` now takes a ``branch_whitelist`` parameter instead of
+  of a ``require_master``
+- ``.travis.yml`` can be used to store some of doctr configuration in addition
+  to the command line flags. Write doctr configuration under the ``doctr`` key.
+  :issue:`137`
+- All boolean command line flags now have a their counterpart that can overwrite
+  the config values set in ``.travis.yml``
 
 1.4.1 (2017-01-11)
 ==================
--- a/docs/commandline.rst
+++ b/docs/commandline.rst
@@ -29,7 +29,7 @@ configuration:
       - cd docs
       - make html
       - cd ..
-      - doctr deploy
+      - doctr deploy .
   - doctr:
       - key-path : 'path/to/key/from/repo/root/path.key'
       - deploy-repo : 'myorg/myrepo'
--- a/docs/commandline.rst
+++ b/docs/commandline.rst
@@ -4,3 +4,42 @@
 
 .. autoprogram:: doctr.__main__:get_parser()
    :prog: doctr
+
+
+Configuration
+-------------
+
+In addition to command line arguments you can configure ``doctr`` using the
+``.travis.yml`` files. Command line arguments take precedence over the value
+present in the configuration file.
+
+The configuration parameters available from the ``.travis.yml`` file mirror
+their command line siblings except doubledashes ``--`` and ``--no-`` prefix are
+dropped.
+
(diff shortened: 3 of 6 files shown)


doctr/__main__.py
```diff
@@ -102,6 +107,16 @@ def get_parser():
 
     return parser
 
+def get_config():
+    """
+    This load some configuration from the `travis.yml`, file is present,
+    `doctr` key if present.
```
asmeurer: Two backquotes for RST.

doctr/__main__.py
```diff
@@ -102,6 +107,16 @@ def get_parser():
 
     return parser
 
+def get_config():
+    """
+    This load some configuration from the `travis.yml`, file is present,
```
asmeurer: .travis.yml

asmeurer: *if

Carreau:
Reviving this as part of the docathon. 

I've cleaned it up a bit and will appreciate review.

Carreau:
Oh, of course test are failing as the PR is from a branch not on this repo.

.travis.yml
```diff
@@ -31,10 +31,14 @@ script:
       cd docs;
       make html;
       cd ..;
-      python -m doctr deploy --gh-pages-docs . --key-path deploy_key.enc;
-      python -m doctr deploy --gh-pages-docs docs --key-path deploy_key.enc;
-      python -m doctr deploy --no-require-master --gh-pages-docs "docs-$TRAVIS_BRANCH" --built-docs docs/_build/html --key-path deploy_key.enc;
-      python -m doctr deploy --no-require-master --key-path deploy_key.enc --no-sync --command "echo test";
+      if [[ -z "${DOCTR_DEPLOY_ENCRYPTION_KEY}" ]]; then
+          echo "DOCTR_DEPLOY_ENCRYPTION_KEY not set, likely building a PR from untrusted repository";
```
asmeurer: We ought to be testing this in doctr itself. Is this happening to you here?

docs/changelog.rst
```diff
@@ -2,6 +2,14 @@
  Doctr Changelog
 =================
 
+Master
+======
+
+- ``setup_GitHub_push`` now takes a ``branch_whitelist`` parameter instead of
+  of a ``require_master``
+- ``.travis.yml`` can be used to store some of doctr configuration in addition
+  to the command line flags. Write doctr configuration under the ``doctr`` key.
```
asmeurer: Add <code>:issue:\`137\`</code> to these.

docs/commandline.rst
```diff
@@ -4,3 +4,37 @@
...
+-------------
+
+In addition to command line arguments you can configure ``doctr`` using the
+``.travis.yml`` files. Command line arguments take precedence over the value
+present in the configuration file.
+
+The configuration parameters available from the ``.travis.yml`` file often
+mirror their command line siblings except dashes ``-`` are replaced by
+underscores ``_``. Some parameters are not available from configuration files
+for security reasons.
```
asmeurer: Can we be more specific here? Are there any values that aren't the same after replacing `-` with `_`? Which flags aren't available for security reasons?

docs/commandline.rst
```diff
@@ -4,3 +4,37 @@
...
+
+Use a ``doctr`` section in your ``travis.yml`` file to store your doctr
+configuration:
+
+.. code::
```
asmeurer: You can set this as YAML language (though IIRC the highlighting might not actually work).

docs/commandline.rst
```diff
@@ -4,3 +4,37 @@
...
+
+  - language: python
+  - script: py.test
+  - doctr:
+    gh_pages_docs: 'documentation'
+    key_path : '/absolute/key/path.key'
```
asmeurer: Absolute key path? The key path should be relative to the github repo.

docs/commandline.rst
```diff
@@ -4,3 +4,37 @@
...
+
+Use a ``doctr`` section in your ``travis.yml`` file to store your doctr
+configuration:
+
+.. code::
+
+  - language: python
+  - script: py.test
+  - doctr:
+    gh_pages_docs: 'documentation'
```
asmeurer: This will change with https://github.com/drdoctr/doctr/pull/157 (which will be merged soon).

.travis.yml
Carreau: Hum, yes on my fork. Let me see if it was not rebased on the lastest.

docs/commandline.rst
```diff
@@ -4,3 +4,37 @@
...
+underscores ``_``. Some parameters are not available from configuration files
+for security reasons.
+
+Use a ``doctr`` section in your ``travis.yml`` file to store your doctr
+configuration:
+
+.. code::
+
+  - language: python
+  - script: py.test
```
asmeurer: You still have to actually call `doctr deploy` in the script though, right?

doctr/__main__.py
```diff
@@ -102,6 +125,20 @@ def get_parser():
...
+    """
+    This load some configuration from the ``.travis.yml``, if file is present,
+    ``doctr`` key if present.
+    """
+    with open('.travis.yml') as travis_config_file:
+        travis_config = yaml.safe_load(travis_config_file.read())
+
+    config = travis_config.get('doctr',{})
+
+    if type(config) == list:
```
asmeurer: `if not isinstance(config, dict)`?

doctr/__main__.py
```diff
@@ -102,6 +125,20 @@ def get_parser():
 
     return parser
 
+def get_config():
+    """
+    This load some configuration from the ``.travis.yml``, if file is present,
+    ``doctr`` key if present.
+    """
+    with open('.travis.yml') as travis_config_file:
```
asmeurer: Add a check if the file exists.

doctr/__main__.py
```diff
@@ -147,7 +191,7 @@ def deploy(args, parser):
             added, removed = [], []
 
         if args.command:
-            run(shlex.split(args.command))
+            run([os.path.expanduser(x) for x in shlex.split(args.command)])
```
asmeurer: Why is this change here? I don't think we should do this.

doctr/__main__.py
```diff
@@ -102,6 +125,20 @@ def get_parser():
...
+    ``doctr`` key if present.
+    """
+    with open('.travis.yml') as travis_config_file:
+        travis_config = yaml.safe_load(travis_config_file.read())
+
+    config = travis_config.get('doctr',{})
+
+    if type(config) == list:
+        raise ValueError('config is not a dict: {}'.format(config))
+    return travis_config.get('doctr',{})
```
asmeurer: `return config`?

.travis.yml
asmeurer: "We ought" meaning we should. I don't know if we actually are (but if we aren't, we need to be). Someone actually reported this same error on SymPy, so I think there may be a real bug here.

In any case, let's fix this in a separate PR.

gforsyth: We do it when we decrypt the key on Travis: https://github.com/drdoctr/doctr/blob/master/doctr/travis.py#L51

asmeurer: That's only supposed to be called with `canpush` is True. https://github.com/drdoctr/doctr/blob/84d09b19b4f045c93f06516dae78c55662b3db26/doctr/travis.py#L185. So either we have a bug or Travis does (maybe `$TRAVIS_PULL_REQUEST` is not always set properly). 

It would be nice if we could keep this as an error, because it is a legitimate error if someone didn't set up their .travis.yml correctly. Maybe there is a more robust way to detect pull requests than looking at the environment variables. 

Let's open an issue for this.

docs/commandline.rst
Carreau: > Can we be more specific here? Are there any values that aren't the same after replacing - with _? Which flags aren't available for security reasons?

Hum, I don't remember. I'm using `safe_load` from Yaml so itm might be due to that. I'll rephrase.

asmeurer: All the parameters are paths (strings). Unexpected stuff can happen in YAML if you don't quote a string and it looks like another literal (like a float), but that's unlikely to happen here. I can't think of any other issues.

Carreau: Oh, yeah. `--token`, if you allow a token option, it means pasting the token in clear in the file.

asmeurer: `--token` is a boolean flag. The token can only be accessed from a encrypted environment variable.

Carreau:
Ok, I'm tired today, I did a lot of mistakes, but that should be all fixed.

asmeurer:
I pushed the branch up to this repo so the tests can run from that.

asmeurer:
We should probably add some configuration to our .travis.yml to test this (the existing tests will still work as they are as the command line flags override the .travis.yml keys).

Carreau:
> We should probably add some configuration to our .travis.yml to test this (the existing tests will still work as they are as the command line flags override the .travis.yml keys).

Done, I had to add opposites of some command line flags like `--sync`/`--no-sync` to be able to overwrite config file.

asmeurer:
Cool. By the way, it's easier to review your changes if you do them as new commits.

Carreau:
> Cool. By the way, it's easier to review your changes if you do them as new commits.

Conflict addressed and push as multiple commits.

asmeurer:
Let's give it a try. Also self reminder we need to add pyyaml to the conda-forge recipe when we release.

PR #141, opened 2016-12-05
add xonsh to readme

scopatz:


asmeurer:
@gforsyth you might want to look at the test failure. Travis is giving a PNG instead of JSON?

gforsyth:
Weeeeirrdddd. So yeah, now if you try to ping an invalid user or repo instead of JSON you get this badge: 
badge

so I guess I can change the test to see if we get JSON back at all, rather than examining the status codes

asmeurer:
Seems like a bug in Travis. Might want to open an issue with them to see if it's a regression. IMO api.* sites shouldn't return non-JSON.

gforsyth:
Thanks, @scopatz!

scopatz:
Sorry it was such a rabbit hole.

gforsyth:
No worries -- it was going to pop up sooner or later!

PR #142, opened 2016-12-05
check if travis returns a PNG instead of json

gforsyth:
Travis has changed how they respond to requests for invalid repos --
returning a PNG badge instead of JSON which makes `json` very sad. While
it's kind of ridiculous, the `check_repo_exists` function now checks to
see if `PNG` is in the content of the request and raises a
`RuntimeError` if so.

I think it's safe to assume that if a PNG has been returned in response
to this call that clearly SOMETHING has gone wrong.


asmeurer:
I would check if this is an upstream bug first (or if we are using the wrong API).

gforsyth:
Closing in favor of #143

PR #143, opened 2016-12-06
ping the correct travis api endpoint

gforsyth:
`repos/{user}/{repo}` used to work but doesn't anymore. But just using
`{user}/{repo}` works well and doesn't randomly return PNGs


asmeurer:
Seems legit if it works.

gforsyth:
There is a possible latent issue with the Travis API, but I don't think it changes the behavior here -- the previous API call is one that's listed in the docs but this should keep working.

Issue #144, opened 2016-12-12
Travis script doesn't stop on failed command


asmeurer:
See https://github.com/travis-ci/travis-ci/issues/1066. 

The result is that if people write their .travis.yml as our example shows, it will run doctr even if the docs fail, which can result in the dev docs being deleted (if the build fails and no docs are generated, it will think that all files should be removed).  

You can fix this either with `set -e` or by putting the command in a yaml block. We should update the docs, and our own .travis.yml.

asmeurer:
The yaml block thing doesn't work. You need to use `set -e`.

PR #146, opened 2016-12-13
add warning about needing `set -e` in travis.yml

gforsyth:
as mentioned in #144.  we were already using `set -e` in doctr's `travis.yml` so that doesn't need any updates.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,8 +68,11 @@ in your ``.travis.yml``. See `the one
 <https://github.com/drdoctr/doctr/blob/master/.travis.yml>`_ used by Doctr
 itself for example.
 
-**Warning: ** To stop a Travis build when the docs fail to build, you must add
- ``set -e`` in ``script``.
+**Warning:** To stop a Travis build when the docs fail to build, you must add
+ ``set -e`` in ``script``. If you put ``doctr deploy`` in ``after_success``
+ instead of ``script``, it will `not cause
+ <https://docs.travis-ci.com/user/customizing-the-build#Breaking-the-Build>`_
+ the build to fail.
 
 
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -65,10 +66,12 @@ something like
 
 in your ``.travis.yml``. See `the one
 <https://github.com/drdoctr/doctr/blob/master/.travis.yml>`_ used by Doctr
-itself for example. Warning, if you put ``doctr deploy`` in ``after_success``
-instead of ``script``, it will `not cause
-<https://docs.travis-ci.com/user/customizing-the-build#Breaking-the-Build>`_
-the build to fail if it fails.
```
asmeurer: This warning shouldn't be deleted.

README.rst
```diff
@@ -65,10 +66,12 @@ something like
...
+itself for example.
+
+**Warning: ** To stop a Travis build when the docs fail to build, you must add
```
asmeurer: I'm guessing Sphinx doesn't like the space here.

asmeurer:
I pushed an update to the wording. I also went ahead and used the `.. warning::` directive. I didn't use it before because it's big and red, but given that both of these gotchas have bitten me, I think it's worth making sure people know about them. 

We should also add some text to the `doctr configure` output.

asmeurer:
Added some text to `doctr configure` output. 

I just realized that maybe not everyone necessarily wants a doctr failure to fail their build, but I think it's good, because it helps us catch bugs (like the original https://github.com/drdoctr/doctr/issues/138 was a bug introduced by a change in Travis CI that wasn't failing the build because they used `after_success`).

asmeurer:
The warnings don't render so nicely on GitHub https://github.com/drdoctr/doctr/tree/sete. I guess it's not a big deal.

PR #147, opened 2016-12-31
specify that git pull should grab doctr_remote

gforsyth:
as discussed in https://github.com/xonsh/xonsh/pull/2039#issuecomment-269878414 
specifies an explicit pull target to avoid grabbing the wrong branch (or even the wrong repo...) still not sure how this happens in the first place but I think this might fix it.

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -379,6 +379,6 @@ def push_docs():
     """
 
     print("Pulling")
-    run(["git", "pull"])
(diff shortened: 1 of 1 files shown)


asmeurer:
I don't know if you can test that this really works before releasing it but it looks good.

gforsyth:
Ok -- I'll merge this in, then try it out on xonsh using a pip install from master and see if it works, then we can put out a new release if it does

PR #152, opened 2017-01-07
Set Travis language to generic

asmeurer:
We don't need Python since we are using Miniconda and this should hopefully be
faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: python
+language: generic
 
 sudo: false
 env:


asmeurer:
Doesn't seem to make a difference for speed 

before: https://travis-ci.org/drdoctr/doctr/builds/189694500 
after: https://travis-ci.org/drdoctr/doctr/builds/189876404

asmeurer:
¯\\\_(ツ)_/¯ ‏

ocefpaf:
Well... Now we know :smile:

PR #154, opened 2017-01-15
Test Python 3.6 on Travis

asmeurer:


asmeurer:
Ah I guess we need to update sphinxcontrib-autoprogram.

asmeurer:
It works now.

Issue #128, opened 2016-10-19
Change --gh-pages-docs default


asmeurer:
It's a backwards incompatible change, but `--gh-pages-docs` defaulting to `docs` is not great. `.` is probably what people expect.  Maybe we should make it a required argument. 

CC @ocefpaf

gforsyth:
I think it's reasonable to change it to `.`

gforsyth:
We could make it a required argument and make the helper text to be pasted into the `.travis.yml` `--gh-pages-docs .` as the default?

ocefpaf:
I do not know the reasons to deploy to `docs`, and I am fine with https://github.com/drdoctr/doctr/issues/128#issuecomment-254960672 solution, but maybe the project is still young and a backwards incompatible change is not that bad...

asmeurer:
If it's required it really ought to be the argument to `doctr deploy`, like `doctr deploy .`.

gforsyth:
Done in #157.

PR #157, opened 2017-02-07
make deploy directory required argument

gforsyth:
as per discussion in #128, changing the deploy directory to a required
argument. the new usage for doctr on the travis side of things would
be

```
doctr deploy .
```

to deploy to the root directory on the ``gh-pages`` branch.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,8 +4,9 @@
 
 Current
 =======
-- There is no longer a default deploy directory. Specify the deploy directory 
-  like ``doctr deploy .`` or ``doctr deploy docs``. (:issue:`128`)
+- The ``--gh-pages-docs`` flag of ``doctr deploy`` has been deprecated.
+  Specify the deploy directory like ``doctr deploy .`` or ``doctr deploy docs``.
+  There is also no longer a default deploy directory. (:issue:`128`)
 
 
 1.4.1 (2017-01-11)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,8 @@
 
 Current
 =======
-- Change deploy directory to required argument. This is a backwards incompatible change. Default deploy without arguments should now read ``doctr deploy .`` (:issue:`128`)
+- There is no longer a default deploy directory. Specify the deploy directory 
+  like ``doctr deploy .`` or ``doctr deploy docs``. (:issue:`128`)
 
 
 1.4.1 (2017-01-11)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,11 @@
  Doctr Changelog
 =================
 
+Current
+=======
+- Change deploy directory to required argument. This is a backwards incompatible change. Default deploy without arguments should now read ``doctr deploy .`` (:issue:`128`)
+
+
 1.4.1 (2017-01-11)
 ==================
 - Fix Travis API endpoint when checking if a repo exists. (:issue:`143`)
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -81,8 +81,10 @@ options available.
         default=True, help="Run all the steps except the last push step."
         "Useful for debugging")
     deploy_parser.add_argument('--gh-pages-docs', default='docs',
-        help="""!!DEPRECATED!! Directory to deploy the html documentation to on gh-pages. The
-        default is %(default)r.""")
+        help="""!!DEPRECATED!! Directory to deploy the html documentation to on gh-pages.
+        The default is %(default)r. The deploy directory should be passed as
+        the first argument to 'doctr deploy'. This flag is kept for backwards
+        compatibility.""")
 
 
     configure_parser = subcommand.add_parser('configure', help="Configure doctr. This command should be run locally (not on Travis).")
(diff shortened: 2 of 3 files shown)


asmeurer:
Let's check all known users of doctr to make sure this doesn't break them.

gforsyth:
I think it will break all users -- I guess we could PR appropriate changes to them?

doctr/__main__.py
```diff
@@ -60,9 +60,8 @@ def get_parser():
     deploy_parser.add_argument('--built-docs', default=None,
         help="""Location of the built html documentation to be deployed to
         gh-pages. If not specified, Doctr will try to automatically detect build location""")
-    deploy_parser.add_argument('--gh-pages-docs', default='docs',
```
asmeurer: Maybe we could keep and deprecate this option for backwards compatibility.

gforsyth: I was thinking of that, but I'm not sure how to go about having a default argument not be required if a certain non-default option is passed.

asmeurer:
I guess we could deprecate it with some custom argparse logic (require the argument if the flag isn't there, and if the flag is used print a deprecation warning). OTOH, deprecation warnings will be ignored, so at best it would just make a transition possible without a period of non-building docs.

doctr/__main__.py
asmeurer: Does argparse let you set a default for positional arguments? You'd have to do the flag logic manually (use `parser.error` to print the error).

gforsyth:
There's definitely a benefit to having a period in which to help people transition, I guess I'll see how difficult it is to implement the custom argparse stuff

asmeurer:
Yeah, even if we PR everyone, they would all have to merge their PRs exactly when we release, or else their docs will fail.

scopatz:
`nargs= '?'`

doctr/__main__.py
scopatz: `nargs= '?'`

asmeurer:
I was thinking it might work by setting a `default` value.  I didn't test it, though.

scopatz:
Well, you have to set a default value, but it will still require the argument to be present. You need both IIRC.

gforsyth:
Ok -- this is ready for another look.  @scopatz was (unsurprisingly) right about `nargs='?'` -- I've also updated `doctr`s `.travis.yml` to use both the deprecated and the new `deploy` format so we can check that both are working for now.

asmeurer:
I pushed a change to the deprecation message.

asmeurer:
Am I missing something, or does this not actually use the deploy_directory variable? It should use it, or `gh_pages_docs` if it isn't there (and neither and both should be an error).

gforsyth:
> Am I missing something, or does this not actually use the deploy_directory variable? It should use it, or gh_pages_docs if it isn't there (and neither and both should be an error).

Nope -- mea culpa.  That's a bit of an oversight.  I'll add that in later tonight.

gforsyth:
This is ready for another look.

docs/changelog.rst
```diff
@@ -2,6 +2,11 @@
  Doctr Changelog
 =================
 
+Current
+=======
+- Change deploy directory to required argument. This is a backwards incompatible change. Default deploy without arguments should now read ``doctr deploy .`` (:issue:`128`)
```
asmeurer: This should be reworded a bit. The default used to be `docs`, but now there is no default. Maybe something like

> There is no longer a default deploy directory. Specify the deploy directory like `doctr deploy .` or `doctr deploy docs`.

asmeurer:
The code looks good. Maybe we should test that it works in Travis? 

Either way, let's check which known repos using doctr need to be updated.

gforsyth:
Updated the changelog.  The list of repos I have using doctr (searching for `doctr` and filtering for YAML files) is:

* [ ] conda
* [ ] sympy
* [ ] xonsh (I can fix this)
* [ ] pygbe (I can fix this)
* [ ] essential_skills_RRC (I can fix this)
* [ ] https://github.com/pydata/pandas-design (although it says deprecated in the repo description)
* [ ] https://github.com/Carreau/warn
* [ ] https://github.com/pyoceans/sea-py
* [ ] https://github.com/maxpoint/spylon
* [ ] https://github.com/danielballan/photomosaic
* [ ] https://github.com/ergs/TWoFCS
* [ ] https://github.com/pystitch/stitch
* [ ] https://github.com/ergs/rg-db-twofcs
* [ ] https://github.com/python-visualization/folium
* [ ] https://github.com/sea-mat/sea-mat
* [ ] https://github.com/bryanwweber/UConnRCMPy
* [ ] https://github.com/bryanwweber/thermostate
* [ ] https://github.com/Alsheh/tmp
* [ ] https://github.com/antepsis/anteplahmacun

asmeurer:
I will fix sympy.

asmeurer:
We should make a new issue about the updates. We can't do it until this is merged and released.

Carreau:
https://github.com/gallantlab/pycortex/pull/205 will switch to doctr (not merged yet) , they should be warned also.

gforsyth:
Ok, anything left on this? We're tracking the people we need to notify about the usage changes.

This will also require a few changes to @Carreaus' #137 if this is merged first, but those should be trivial.

Carreau:
> Ok, anything left on this? We're tracking the people we need to notify about the usage changes.

+1

> This will also require a few changes to @Carreaus' #137 if this is merged first, but those should be trivial.

No problem, I can take care of that.

asmeurer:
I think we should try to get the other PRs in before we release. But this looks good.

Issue #104, opened 2016-08-31
Automatically edit the .travis.yml


asmeurer:
It would be nice if it automatically added the encrypted environment variable to .travis.yml. It's probably not possible to be smart enough to add `doctr deploy`. 

I don't know how to do this well. Probably you need ruamel.yaml to round-trip parse the yaml.

PR #158, opened 2017-02-08
Update travis yml

gforsyth:
An attempt to address #104 -- this was harder than I thought.

I tried to cover all the situations I could think of re: which fields might have already been defined or not in someone's `.travis.yml`. 

One issue (which may be a deal breaker) it doesn't preserve newlines for multiline yaml blocks -- I have it configured so that when those lines are rewritten (after round trip parsing) that they are still valid (basically, put them all on one line) but maybe it's better to warn people before we do that and ask them if it's ok?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
   - conda config --add channels conda-forge # For sphinxcontrib.autoprogram
   - conda update -q conda
   - conda info -a
-  - conda create -q -n test-environment python=3.5 requests cryptography sphinx pyflakes sphinxcontrib-autoprogram pytest sphinx-issues
+  - conda create -q -n test-environment python=3.5 requests cryptography sphinx pyflakes sphinxcontrib-autoprogram pytest sphinx-issues ruamel.yaml
   - source activate test-environment
 
 script:
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -30,7 +30,8 @@ import subprocess
 from textwrap import dedent
 
 from .local import (generate_GitHub_token, encrypt_variable, encrypt_file,
-    upload_GitHub_deploy_key, generate_ssh_key, check_repo_exists, GitHub_login)
+    upload_GitHub_deploy_key, generate_ssh_key, check_repo_exists, GitHub_login,
+    update_travis_yml)
 from .travis import (setup_GitHub_push, commit_docs, push_docs,
     get_current_repo, sync_from_log, find_sphinx_build_dir, run)
 from . import __version__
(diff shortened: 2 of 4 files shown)


asmeurer:
What's the deal with the newlines? The round-trip parser doesn't handle them?

asmeurer:
Looks exciting. Maybe we could test it out on some real life .travis.ymls. [SymPy's](https://github.com/sympy/sympy/blob/master/.travis.yml) is pretty complicated.

asmeurer:
What does this do if there's already an encrypted variable in the file?

gforsyth:
The round trip parser doesn't seem to recognize semi-colons as valid line endings -- or rather, it correctly recognizes a semi-colon separated line or lines as being a single yaml directive, so it groups them together.  When it writes back to a file, it doesn't preserve the newlines after the semicolons, but just writes them all to one line -- so, for instance, the `if DOCS` logic in our `.travis.yml` will be condensed to a single long line. 

I'm curious to see how it handles SymPy's travis.yml -- if that works, then I'm comfortable with the logic (pending some kind of warning about the newline removal).

If there's already a an encrypted variable, this will just add another one after it, so something like

```console
env:
  global:
    - secure: gobbledy-gook
    - secure: more-gobbledy-gook
```

If it's the _same_ encrypted variable name, I imagine the second one will overwrite the first?  I'm not sure about that but it shouldn't be too difficult to confirm that.

asmeurer:
Well often (but not always) an existing secure variable means that the old one was a previous doctr environment variable (there's no way to tell). So we should maybe warn the user. Also, since the variable is gobbleddy-gook, should we add a comment stating what the variable is?

gforsyth:
Ok -- if there's already a secure variable in the .travis.yml then a warning is issued.  Something to the effect that we're appending a new one to the list of global secure variables and this could potentially overwrite something in the environment.
Also now adds a comment with the current date when the key is added so people know which key belongs to doctr (from now on).

I tested it on sympy's `.travis.yml` and that exposed some issues which I've fixed regarding block indentations etc. Now I think it works, even on that monster. Note, though, that there's a significant diff when I run it on sympy's travis.yml since a bunch of the indentation is changed (although not the physical blocks, just the spacing). This could potentially be annoying, but it shouldn't _break_ anything.

asmeurer:
Yeah, that is annoying. I guess ruamel isn't as round-trippy as we would have hoped?

gforsyth:
Apparently not? The documentation is well nigh nonexistent, so there's possibly some way to do it that I've missed.  I'm not quite ready to give up on this -- at the very least, we could offer to do it with a warning that it might screw around with indentation

asmeurer:
Well it's your call, but personally I think adding a bunch of random whitespace to .travis.yml would be very annoying. As a user, I'd probably rather it just not try to edit it at all if it does that.

gforsyth:
I'm going to close this out since the round-trip parser doesn't seem to be in a state to do this cleanly.

asmeurer:
That's too bad. I'd really like to see this feature.

Issue #162, opened 2017-03-07
doctr configure instruction.


Carreau:
Feedback from user using it for the first time:

- remind that doctr-deploy should be run from root of repo (or it may not find the encrypted key).

gforsyth:
Thanks @Carreau ! Yeah, I've actually forgotten this myself a time or two. It shouldn't be too hard to detect if we're in a git repo and if so, to then warn if we aren't at the root of it.

PR #163, opened 2017-03-08
Check root

gforsyth:
If we are in a git repo but not at the root of it, print error and allow
user to ``--force`` doctr to run anyway.

If not in a git repo at all, raise a RuntimeError

Should address #162

### DIFF
--- a/doctr/local.py
+++ b/doctr/local.py
@@ -248,3 +248,20 @@ def check_repo_exists(deploy_repo, service='github', *, auth=None, headers=None)
     r.raise_for_status()
 
     return r.json().get('private', False)
+
+def in_git_root():
+    """
+    Check if current working directory is root dir of git repo.
+    Return ``False`` if it isn't.
+    """
+    repo = subprocess.run(['git', 'rev-parse', '--show-toplevel'],
+                          stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    if repo.returncode == 0:
+        root_dir = repo.stdout.decode('utf-8').strip()
+        if os.getcwd() == root_dir:
+            return True
+        else:
+            return False
+    else:
+        raise RuntimeError('doctr should be run in the root directory of a '
+                           'git repository')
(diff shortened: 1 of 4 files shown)


doctr/local.py
```diff
@@ -248,3 +248,20 @@ def check_repo_exists(deploy_repo, service='github', *, auth=None, headers=None)
     r.raise_for_status()
 
     return r.json().get('private', False)
+
+def in_git_root():
+    """
+    Check if current working directory is root dir of git repo.
+    Return ``False`` if it isn't.
+    """
+    repo = subprocess.run(['git', 'rev-parse', '--show-toplevel'],
```
asmeurer: Can't you just check if the current directory has a `.git` directory?

asmeurer: That wouldn't check if we are in a git repo at all, but I don't see the point have having the distinct error messages in that case.

asmeurer:
The tests are failing from that race condition again.

doctr/local.py
gforsyth: My thought was that someone _might_ want to run `doctr` in a subdirectory for some specific reason? I can't think of what that would be, but that was the thought.

asmeurer: Well it doesn't have to be at the root of the repo, unless I'm mistaken, as long as the paths to the key and the other arguments are given to it, relative to the current directory. It does need to be in the repo, as it calls git. 

So I actually think it should only error when it is not in the git repo (it will error anyway, eventually, because git will complain as soon as doctr calls it), and warn, but not error when it isn't, just so it is easier for people to debug if they accidentally are cd'd to the wrong directory and are trying to figure out why doctr failed.

asmeurer: OTOH, maybe weird stuff will happen when checking out gh-pages if we aren't in the root (unless the directory we are in isn't tracked, or is in both branches).

ylemkimon:
How about checking current working directory is TRAVIS_BUILD_DIR?

asmeurer:
This is for `doctr configure`, which is run on the local compute, not Travis.

ylemkimon:
@asmeurer I thought this was for #274. I think checking current working directory is `TRAVIS_BUILD_DIR` may address it.

PR #170, opened 2017-03-17
only allow pushes from master or whitelisted branches

gforsyth:
If I'm right, then we should know that this is the fix since it won't push to the `gh-pages` branch

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -151,7 +151,7 @@ def setup_GitHub_push(deploy_repo, auth_type='deploy_key', full_key_path='github
 
     if require_master is not None:
         import warnings
-        warnings.warn("`setup_GitHub_push`'s `require_master` argument in favor of `branch_whitelist=['master']`", 
+        warnings.warn("`setup_GitHub_push`'s `require_master` argument in favor of `branch_whitelist=['master']`",
                 DeprecationWarning,
                 stacklevel=2)
         branch_whitelist.add('master')
@@ -163,7 +163,7 @@ def setup_GitHub_push(deploy_repo, auth_type='deploy_key', full_key_path='github
     TRAVIS_BRANCH = os.environ.get("TRAVIS_BRANCH", "")
     TRAVIS_PULL_REQUEST = os.environ.get("TRAVIS_PULL_REQUEST", "")
 
-    if any([re.compile(x).match(TRAVIS_BRANCH) for x in branch_whitelist]):
+    if not any([re.compile(x).match(TRAVIS_BRANCH) for x in branch_whitelist]):
         print("The docs are only pushed to gh-pages from master. To allow pushing from "
         "a non-master branch, use the --no-require-master flag", file=sys.stderr)
         print("This is the {TRAVIS_BRANCH} branch".format(TRAVIS_BRANCH=TRAVIS_BRANCH), file=sys.stderr)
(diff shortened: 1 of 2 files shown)


gforsyth:
Ok, this worked and I just tested on one of my repos, too. Once the merge commit clears CI I'll merge this and then do a new release

gforsyth:
Ok, fool me once, etc...
I broke out the `canpush` logic into its own function and added some tests to make sure this doesn't happen again.  Also, as per @Carreau 's suggestion, changed `if branch_whitelist is None` to `if not branch_whitelist` in order to catch possibility that the whitelist is an empty set or list

doctr/travis.py
```diff
@@ -146,32 +146,20 @@ def setup_GitHub_push(deploy_repo, auth_type='deploy_key', full_key_path='github
     For ``auth_type='deploy_key'``, this sets up the remote with ssh access.
     """
 
-    if branch_whitelist is None:
+    if not branch_whitelist:
         branch_whitelist={'master'}
 
     if require_master is not None:
```
Carreau: do you want to actually remove the `require_master ` param here, it is not used anymore. 
If you are not comiting to Python api that would be a good thing.

doctr/travis.py
```diff
@@ -401,3 +389,22 @@ def push_docs(deploy_branch='gh-pages'):
     run(['git', 'pull', 'doctr_remote', deploy_branch])
     print("Pushing commit")
     run(['git', 'push', '-q', 'doctr_remote', deploy_branch])
+
+def determine_push_rights(branch_whitelist):
+
+    canpush = True
+
+    TRAVIS_BRANCH = os.environ.get("TRAVIS_BRANCH", "")
+    TRAVIS_PULL_REQUEST = os.environ.get("TRAVIS_PULL_REQUEST", "")
```
Carreau: I would have made `TRAVIS_BRANCH` and `TRAVIS_PULL_REQUEST ` parameters.

gforsyth:
Oh. Yeah. In fact I think I have to do that or it'll break things

doctr/travis.py
```diff
@@ -401,3 +389,22 @@ def push_docs(deploy_branch='gh-pages'):
     run(['git', 'pull', 'doctr_remote', deploy_branch])
     print("Pushing commit")
     run(['git', 'push', '-q', 'doctr_remote', deploy_branch])
+
+def determine_push_rights(branch_whitelist):
```
Carreau: Should we test that `branch_whitelist` is a container (or at least not a string?)

doctr/travis.py
gforsyth: I don't think anyone is using the Python API -- let's remove it

gforsyth: But maybe we can do that in a different PR?  I'd like to get this in before people who have `pip install doctr` start experiencing pain

doctr/travis.py
gforsyth: yes, but let's do that separately -- I think this is good enough for now, since the alternative is broken functionality

doctr/travis.py
asmeurer: I checked on GitHub and there are no `import doctr`s anywhere. I was thinking of using the Python API for a project, but I haven't implemented it yet. We shouldn't be too worried about changing it I'd say.

Carreau:
> But maybe we can do that in a different PR? I'd like to get this in before people who have pip install doctr start experiencing pain

yes, as soon as travis is green, we can merge.

gforsyth:
Starting the release now

Carreau:
thanks.

PR #172, opened 2017-03-18
Deprecate --no-require-master in favor of branch-whitelist.

Carreau:
it still seem to be relatively complicated to pass a list as a CLI flag,
so `.travis.yml` only.

I can also have --no-require-master still exist and add `.*` to the
whitelist.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,10 +31,10 @@ script:
       cd docs;
       make html;
       cd ..;
-      python -m doctr deploy --sync .;
-      python -m doctr deploy --sync --gh-pages-docs docs;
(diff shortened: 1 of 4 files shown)


asmeurer:
I don't see how it's complicated to add a list. Just use `nargs='*'`.  If it has special characters at the command line you may need to single quote it, but that's it.

asmeurer:
We now have --branch-whitelist which accepts a list. I like --no-require-master because it's the most common case. It also isn't strictly possible with --branch-whitelist unless we add regex support. 

So I'm closing this for now.

PR #176, opened 2017-03-20
[Prototype] Allow users to deploy on their own fork/PR

Carreau:
This is only the first half (and still draft), instead of commiting the
encryption key, it base64 encode it and set it on travis-ci using the
API.

The second part would be to detect that the env variable is set and
trigger the right behavior, which imho is to deploy the current branch
on `<user-fork>:gh-pages` branch, under a `docs-$BRANCHNAME` subfolder.
WE can't do `$PRNUMBER` as branch will be deployed from the users run of
travis on not from the PR one.

-- 
There is quite a bit of refactoring and code sharing to do, but not today.

### DIFF
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -35,7 +35,7 @@ from pathlib import Path
 from textwrap import dedent
 
 from .local import (generate_GitHub_token, encrypt_variable, encrypt_file,
-    upload_GitHub_deploy_key, generate_ssh_key, check_repo_exists, GitHub_login)
+    upload_GitHub_deploy_key, generate_ssh_key, check_repo_exists, GitHub_login, set_encrypted_variable)
 from .travis import (setup_GitHub_push, commit_docs, push_docs,
     get_current_repo, sync_from_log, find_sphinx_build_dir, run)
 from . import __version__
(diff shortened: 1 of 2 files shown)


doctr/__main__.py
```diff
@@ -122,6 +122,20 @@ def get_parser(config=None):
     deploy_parser = subcommand.add_parser('deploy', help="""Deploy the docs to GitHub from Travis.""")
     deploy_parser.set_defaults(func=deploy)
     deploy_parser_add_argument = make_parser_with_config_adder(deploy_parser, config)
+
+    _user_fork_parser = subcommand.add_parser('fork-deploy', help="""Deploy the docs to GitHub from Travis to the current user fork""")
```
asmeurer: Shouldn't it use the same command (`doctr deploy`), so that users can enable it and have it just work with the same .travis.yml as the main repo?

doctr/__main__.py
```diff
@@ -388,6 +402,74 @@ def configure(args, parser):
...
+        deploy_repo = build_repo
+
+    if deploy_repo != build_repo:
+        check_repo_exists(deploy_repo, service='github', **login_kwargs)
+
+    N = IncrementingInt(1)
+
+    header = "\n================== You should now do the following ==================\n"
+
+    if args.token:
```
asmeurer: You can leave this not implemented if you want. Tokens are less secure (they grant access to all the user's repos), and should really only be used for dedicated bot users (which won't be the case here).

doctr/__main__.py
Carreau: No, it could also be doctr-configure, but this command have to be run on user machine. 
I'm planning on having `doctr deploy` do automatically the right things is the USER_DEPLOY env variable is set.

asmeurer: Oh, then it should be called `doctr configure-fork` or something like that. Right now this looks like the `deploy` command, which gets run on Travis only.

doctr/__main__.py
Carreau: Yes, thanks,  I'll clean things up, I'm just prototyping by copy-pasting for now.

doctr/__main__.py
asmeurer: Or maybe just `doctr configure --fork`.

PR #177, opened 2017-03-29
re-enable pushing from non-master branches

gforsyth:
This was raised by @jorisvandenbossche -- it looks like some part of the `branch_whitelist` additions removed the functionality provided by `--no-require-master`

I know that that's a flag up for deprecation, but it should at least work while we still offer it

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -133,6 +133,20 @@ def get_current_repo():
     _, org, git_repo = remote_url.rsplit('.git', 1)[0].rsplit('/', 2)
     return (org + '/' + git_repo)
 
+def get_travis_branch():
+    """Get the name of the branch that the PR is from.
+
+    Note that this is not simply ``$TRAVIS_BRANCH``. the ``push`` build will
+    use the correct branch (the branch that the PR is from) but the ``pr``
+    build will use the _target_ of the PR (usually master). So instead, we ask
+    for ``$TRAVIS_PULL_REQUEST_BRANCH`` if it's a PR build, and
+    ``$TRAVIS_BRANCH`` if it's a push build.
+    """
+    if os.environ.get("TRAVIS_PULL_REQUEST", "") == "true":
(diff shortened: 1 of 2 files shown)


asmeurer:
Why did we want to deprecate it again? It effectively sets a whitelist of all branches, and I think most people will want either just master or all branches.

gforsyth:
We haven't deprecated it -- it's just been proposed in #172

asmeurer:
Looks like it is working https://drdoctr.github.io/doctr/docs-push_from_branch/api.html#doctr.travis.get_travis_branch

gforsyth:
I'm fine with leaving both options in

asmeurer:
I see. Whitelisting all branches from the command line is more difficult. Even if you use a regular expression, you have to make sure to quote it correctly at the command line, so I think just a simple flag that does it makes sense.

doctr/travis.py
```diff
@@ -133,6 +133,20 @@ def get_current_repo():
...
+    """Get the name of the branch that the PR is from.
+
+    Note that this is not simply ``$TRAVIS_BRANCH``. the ``push`` build will
+    use the correct branch (the branch that the PR is from) but the ``pr``
+    build will use the _target_ of the PR (usually master). So instead, we ask
+    for ``$TRAVIS_PULL_REQUEST_BRANCH`` if it's a PR build, and
+    ``$TRAVIS_BRANCH`` if it's a push build.
+    """
+    if os.environ.get("TRAVIS_PULL_REQUEST", "") == "true":
+        return os.environ.get("TRAVIS_PULL_REQUEST_BRANCH", "")
```
asmeurer: Is this the same logic it used before this was broken?

gforsyth: Honestly, I don't remember. It might've just been using `TRAVIS_BRANCH` -- I didn't discover until earlier today that that doesn't always mean what I thought it meant

gforsyth:
Thanks for looking it over, @asmeurer !

PR #187, opened 2017-04-12
skip github tests if no API token present

gforsyth:
Just a simple fix to skip the tests that ping github if there's no API token present


asmeurer:
We just need to remember to push up a branch for any PR that might affect this.

PR #190, opened 2017-05-22
ENH: Support pushing to *.github.io from other repo.

danielballan:
Doctr already supports publishing to the master branch of an `org/org.github.io` repo. I believe it does not support pushing from `org/some_repo` (and perhaps multiple other repos) into `org/org.github.io`. (Why do this? Different projects in an org can publish into subdirectories of one org-wide gh-pages repo and get a nice URL structure: org.github.io/repo_a, org.github.io/repo_b, etc. You could do this with separate gh-pages branches, but it's nice to keep the large build projects off in their own org-wide repo and keep the repos' branches light. The projects matplotlib, cmocean, and trackpy all do this.)

The issue arises from `deploy_branch` being`master`. Local `master` refers to the history of `org/some_repo`, while `doctr_remote/master` refers to the history of `org/org.github.io` and doctr mixes them up.

I have resolved this by doing the local work on a special branch with an unambiguous name that tracks `doctr_remote/{deploy_branch}` and is explicitly pushed to it.

I *think* this modification does not break existing use cases but someone more familiar with the codebase should take a careful look. This [Travis build](https://travis-ci.org/NSLS-II/docs/builds/234984797) shows it operating successfully for my use case of off this PR.

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -15,6 +15,8 @@ import time
 
 from cryptography.fernet import Fernet
 
+DOCTR_WORKING_BRANCH = '__doctr_working_branch'
+
 def decrypt_file(file, key):
     """
     Decrypts the file ``file``.
@@ -242,13 +244,9 @@ def checkout_deploy_branch(deploy_branch, canpush=True):
     Checkout the deploy branch, creating it if it doesn't exist.
     """
     #create empty branch with .nojekyll if it doesn't already exist
-    new_deploy_branch = create_deploy_branch(deploy_branch, push=canpush)
-    print("Checking out {}".format(deploy_branch))
-    local_deploy_branch_exists = deploy_branch in subprocess.check_output(['git', 'branch']).decode('utf-8').split()
-    if new_deploy_branch or local_deploy_branch_exists:
-        run(['git', 'checkout', deploy_branch])
-    else:
-        run(['git', 'checkout', '-b', deploy_branch, '--track', 'doctr_remote/{}'.format(deploy_branch)])
+    create_deploy_branch(deploy_branch, push=canpush)
+    print("Checking out doctr working branch tracking doctr_remote/{}".format(deploy_branch))
+    run(['git', 'checkout', '-B', DOCTR_WORKING_BRANCH, '--track', 'doctr_remote/{}'.format(deploy_branch)])
     print("Done")
 
 def deploy_branch_exists(deploy_branch):
@@ -448,7 +446,7 @@ def push_docs(deploy_branch='gh-pages', retries=3):
         print("Pulling")
         code = run(['git', 'pull', 'doctr_remote', deploy_branch])
         print("Pushing commit")
-        code = run(['git', 'push', '-q', 'doctr_remote', deploy_branch], exit=False)
+        code = run(['git', 'push', '-q', 'doctr_remote', '{}:{}'.format(DOCTR_WORKING_BRANCH, deploy_branch)])
         if code:
             retries -= 1
             print("Push failed, retrying")
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -246,8 +246,16 @@ def checkout_deploy_branch(deploy_branch, canpush=True):
     #create empty branch with .nojekyll if it doesn't already exist
     create_deploy_branch(deploy_branch, push=canpush)
     print("Checking out doctr working branch tracking doctr_remote/{}".format(deploy_branch))
-    run(['git', 'checkout', '-B', DOCTR_WORKING_BRANCH, '--track', 'doctr_remote/{}'.format(deploy_branch)])
+    clear_working_branch()
+    run(['git', 'checkout', '-b', DOCTR_WORKING_BRANCH, '--track', 'doctr_remote/{}'.format(deploy_branch)])
     print("Done")
+
+    return canpush
+
+def clear_working_branch():
+    local_branch_names = subprocess.check_output(['git', 'branch']).decode('utf-8').split()
+    if DOCTR_WORKING_BRANCH in local_branch_names:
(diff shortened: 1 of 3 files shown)


asmeurer:
Tests fail (pyflakes).

I'll push the branch up to origin so that it tests the deploy (by the way, this is the best branch name I've ever seen).

doctr/travis.py
```diff
@@ -214,12 +216,8 @@ def setup_GitHub_push(deploy_repo, auth_type='deploy_key', full_key_path='github
 
     #create empty branch with .nojekyll if it doesn't already exist
     new_deploy_branch = create_deploy_branch(deploy_branch, push=canpush)
-    print("Checking out {}".format(deploy_branch))
-    local_deploy_branch_exists = deploy_branch in subprocess.check_output(['git', 'branch']).decode('utf-8').split()
-    if new_deploy_branch or local_deploy_branch_exists:
-        run(['git', 'checkout', deploy_branch])
-    else:
-        run(['git', 'checkout', '-b', deploy_branch, '--track', 'doctr_remote/{}'.format(deploy_branch)])
```
asmeurer: Does this still automatically create gh-pages on the remote if it doesn't exist?

asmeurer:
Seems like this would have needed to have been done in the first place for .github.io support to be working.

asmeurer:
I'd like to see a deploy added to .travis.yml to deploy to drdoctr.github.io (it seems that wasn't done in https://github.com/drdoctr/doctr/pull/168/).

doctr/travis.py
danielballan: Yes, if I understand correctly that part happens in create_deploy_branch which is still called, just above the first line I touched. To be clear I have not tested that; I just don't think the behavior had changed.

danielballan:
> Seems like this would have needed to have been done in the first place for .github.io support to be working.

Seemed that way to me too but I figured someone would have caught it by now? I'll try to add that test you suggest.

danielballan:
I amended my original commit to fix the pyflakes failure and added a deployment to drdoctr.github.io.

doctr/travis.py
danielballan: Actually I was wrong about that -- it did not successfully create a `master` branch if one did not exist -- but that is fixed in 8007280, which I think will pass once it runs in context where it's allowed to push.

danielballan:
I think this will pass if it runs on the upstream fork.

danielballan:
(The failure is: we can't create a branch that tracks `doctr_remote/master` if it doesn't exist, and we can't create it if `canpush` is False.)

danielballan:
Please let me know when you have the bandwidth to review/merge and I will happily rebase.

asmeurer:
I can review, but I think this PR was not working (at least it was failing the tests).

asmeurer:
Now it failed:

```
git fetch doctr_remote
Checking out doctr working branch tracking doctr_remote/master
git checkout -b __doctr_working_branch --track doctr_remote/master
fatal: Cannot update paths and switch to branch '__doctr_working_branch' at the same time.
Did you intend to checkout 'doctr_remote/master' which can not be resolved as commit?
```

By the way, we should put an except block in our `try/finally` that prints an exception message to make it easier to find the actual error in the build logs.

danielballan:
Thanks. The ball is in my court then. I'll see if I can get the tests to pass.

danielballan:
@asmeurer It looks like the failing builds (the ones in travis-ci/pr) haven't run since yesterday, before the master branch was created. Would you re-run them? I don't think I have the necessary permissions to do that.

danielballan:
Ah, pushing a new (empty) commit made Travis spin. Tests pass. Now let me remove that empty commit....

asmeurer:
I did a merge on master after merging another PR, which should make it re-run. I'll merge this is Travis passes.

danielballan:
What sort of error message did you have in mind for that try/finally block? Do you want something with caps or whitespace to stand out in the logs? Or is there something slicker we can do?

asmeurer:
I'm working on the message at https://github.com/drdoctr/doctr/pull/239. If you have a better suggestion let me know.

asmeurer:
OK, let's see if https://github.com/drdoctr/drdoctr.github.io gets updated.

asmeurer:
Oh I need to run doctr configure and actually set up the deploy key. But why didn't the build fail? https://travis-ci.org/drdoctr/doctr/jobs/267309777

asmeurer:
I think maybe it's because we already have `doctr_remote` from the other doctr commands. It too should be cleaned up in the finally block.

asmeurer:
Oh nevermind that's already taken care of. The docs aren't pushed because it thinks they haven't changed.

asmeurer:
OK, I think you need to clear the `__doctr_working_branch` in the finally block. I'll open a PR.

asmeurer:
No, you're already doing that too. I'm not seeing why it's failing.

asmeurer:
Oh, I see. It's because we (probably stupidly) have it set up in our Travis that you have to have `--sync` for the docs to actually push.

asmeurer:
Which actually leads to a second question, which is how we can actually test this, since doctr is currently set up to use only one deploy key, but we would need two. 

I guess I can generate a new deploy key manually and use it for both repos (hopefully GitHub allows that).

asmeurer:
No, GitHub doesn't allow it.

danielballan:
Right. Is it worth adding something to the test that makes a tiny random change to the docs each time?

asmeurer:
The docs do get updated each time (sphinx adds timestamps). So that isn't necessary. I added --sync [here](https://github.com/drdoctr/doctr/pull/241), but it won't work until we can get a deploy key set up, which we need to figure out how to do (https://github.com/drdoctr/doctr/issues/242).

danielballan:
Ah, got it.

PR #192, opened 2017-05-31
Run --command on the original branch, not the deploy branch

asmeurer:
Fixes https://github.com/drdoctr/doctr/issues/191

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ script:
       python -m doctr deploy --sync .;
       python -m doctr deploy --sync --gh-pages-docs docs;
       python -m doctr deploy --sync --no-require-master  --built-docs docs/_build/html "docs-$TRAVIS_BRANCH";
-      python -m doctr deploy --no-require-master  --command "echo test" docs;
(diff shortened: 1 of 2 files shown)


asmeurer:
Looks like it is working.

gforsyth:
Looks good, sorry for the delay @asmeurer

asmeurer:
This actually breaks the SymPy use of doctr, which relies on running a script in the deploy repo (separate from the main repo). Going to revert for now and do a bugfix release. I didn't end up needing my original use for this, so I'll probably just leave it as an open issue to be able to run --command in the original repo.

asmeurer:
Actually a little complicate to do this. I'm going to try a workaround where I pull in the script from `doctr_remote/gh-pages` using gh-pages before running it.  If it works, we should document it.

asmeurer:
Oh that doesn't work. It needs the files from the deploy repo. I guess we should have a --pre-command and a --post-command.

asmeurer:
OK, I actually don't think a pre-checkout command (what --command does in 1.6) is useful. If you want to run a command before anything, just run it in the build before you run doctr.

PR #193, opened 2017-05-31
Run --command with shell=True

asmeurer:
This is what most users will expect, and without this running more than one
command would require creating a script file.

Will have merge conflicts once https://github.com/drdoctr/doctr/pull/192 is merged.


asmeurer:
Looks like it is working. I wanna do a release once this and #192 are in.

gforsyth:
Conflicts on this now

asmeurer:
Merged in master, and double checked that it still works.

gforsyth:
Looks good.  Thanks, @asmeurer !

PR #196, opened 2017-06-13
Remove Sphinx stuff from main page

### LINKED ISSUE #195: People still think doctr is Sphinx only
Apparently people still think doctr is Sphinx only. We should figure out how to make this clearer in the docs. Maybe remove mentions of Sphinx from the main page. CC @bjodah.

asmeurer:
Fixes #195.


gforsyth:
Yep, looks good. Hopefully it helps with the confusion.

PR #203, opened 2017-06-21
call to authorization api to generate 2FA

### LINKED ISSUE #197: two factor authentication not sending 2FA code
I have been trying to get 2FA auth code but it just does not work. I had to get 2FA code from the github.com website.

Something like below seems to send me 2FA code:

```shell
curl -u techgaun -d '{"scopes": ["user:email"], "note": "give me 2FA"}' https://api.github.com/authorizations
```

but the following does not:
```shell
curl -u techgaun https://api.github.com/
```

So seems like its github's API that really needs fixed but I'm going to open an issue here as this has been how the github API behaves. Funny thing is that when I go and check logs after doing `doctr configure`, I can see the two_factor_requested in the log but I never get the 2FA code (I'm using SMS). I have to make the explicit call like the first one to get 2FA code

techgaun:
given it requires 2FA, the generate token call is going to fail but thats going to send 2FA code as expected.. trades off one more api call for better UX.. closes #197

### DIFF
--- a/doctr/local.py
+++ b/doctr/local.py
@@ -128,6 +128,12 @@ def GitHub_login(*, username=None, password=None, OTP=None, headers=None):
     if r.status_code == 401:
         two_factor = r.headers.get('X-GitHub-OTP')
         if two_factor:
+            auth_header = base64.urlsafe_b64encode(bytes(username + ':' + password, 'utf8')).decode()
+            login_kwargs = {'auth': None, 'headers': {'Authorization': 'Basic {}'.format(auth_header)}}
+            try:
+                generate_GitHub_token(**login_kwargs)
+            except requests.exceptions.HTTPError:
+                pass
             print("A two-factor authentication code is required:", two_factor.split(';')[1].strip())
             OTP = input("Authentication code: ")
             return GitHub_login(username=username, password=password, OTP=OTP, headers=headers)


asmeurer:
It (still) works for me, with app 2FA. If you can confirm it works for text 2FA we can merge.

techgaun:
@asmeurer yeah it works fine with text 2FA.. I've used bunch of times since last couple of days

asmeurer:
Thanks.

doctr/local.py
```diff
@@ -128,6 +128,12 @@ def GitHub_login(*, username=None, password=None, OTP=None, headers=None):
     if r.status_code == 401:
         two_factor = r.headers.get('X-GitHub-OTP')
         if two_factor:
+            auth_header = base64.urlsafe_b64encode(bytes(username + ':' + password, 'utf8')).decode()
+            login_kwargs = {'auth': None, 'headers': {'Authorization': 'Basic {}'.format(auth_header)}}
```
asmeurer: Should this be passing in the additional headers passed in to the function?

techgaun: it could potentially but its not really necessary to generate SMS so I just used it like this

asmeurer: As far as I can tell nothing actually uses this keyword arg other than the recursive calls in this function, so let's not worry about it.

PR #204, opened 2017-06-21
fix travis pr check

### LINKED ISSUE #199: get_travis_branch() giving wrong branch for PR
`get_travis_branch()` says it gives the right branch for a PR, but it seems to be giving master for me. This is on a PR from a branch on the same repo.

techgaun:
based on the docs : https://docs.travis-ci.com/user/environment-variables/

TRAVIS_PULL_REQUEST: The pull request number if the current job is a pull request, “false” if it’s not a pull request.

This closes #199

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -148,7 +148,7 @@ def get_travis_branch():
     for ``$TRAVIS_PULL_REQUEST_BRANCH`` if it's a PR build, and
     ``$TRAVIS_BRANCH`` if it's a push build.
     """
-    if os.environ.get("TRAVIS_PULL_REQUEST", "") == "true":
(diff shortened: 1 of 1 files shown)


asmeurer:
I think the test failure is relevant here.

techgaun:
oh crap I will check it out @asmeurer

techgaun:
@asmeurer seems like this requires some more thought as I don't think https://github.com/drdoctr/doctr/blob/421aefdc0246713727f50d17669ff4f0efda684c/doctr/__main__.py#L270 would work. It would work for branch within the same repo but not on a different repo (like my fork for eg.). What do you suggest in this scenario?

asmeurer:
Hmm. It seems really that should be use the merge commit generated by Travis, not the original branch.

PR #219, opened 2017-07-15
Stash instead of using git reset --hard

### LINKED ISSUE #218: Corner cases for calling doctr multiple times
After https://github.com/drdoctr/doctr/pull/215, we call git reset --hard, so if you call doctr a second time, and want it to deploy a file tracked in master that was modified, it won't be modified in the same way. 

Maybe a better way would be to stash and pop.

asmeurer:
That way untracked changes are left intact for any additional commands, such
as multiple doctr calls.

Fixes #218.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,8 @@ install:
 script:
   - set -e
   - set -x
-  - if [[ "${DOCS}" == "true" ]]; then
+  - |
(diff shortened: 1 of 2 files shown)


asmeurer:
The way I set this up I can't really test it until I merge.

PR #225, opened 2017-07-16
Allow building from tags

### LINKED ISSUE #29: Ability to do something different on a tag
This tool is mainly designed for uploading dev docs, but it can also be used to upload release docs. In that case, we might need some way to do something different on a tag. 

Or maybe the logic for uploading on a tag should just go in the individual .travis.yml files, in which case we can at least document what it could look like.

asmeurer:
Adds --build-tags flag.  How to use is described in the recipes in the docs.

Will need to push some tags to GitHub to test this.

Fixes #29.


asmeurer:
My thinking here is that we can use the same `doctr deploy` command for both tags and branch builds, with different deploy directories handled with logic outside of the command. 

But now I'm wondering what's the cleanest way to deploy only from tags. There's no way to specify the branch whitelist from the command line.

asmeurer:
Small compatibility break with `--branch-whitelist` here. Needs to be documented in the release notes.

asmeurer:
OK, I think the code is ready, but we need to test it by pushing some tags. Probably need to merge first.

asmeurer:
Ended up being a little fragile. We should test this pretty hard once merging.

CJ-Wright:
Is this ready to merge?

asmeurer:
I think it needs some more testing. Maybe I should push some fake tags up for this branch to make sure it all works.

asmeurer:
I think I might hold off on merging this until after the next release (1.6), since there are already a lot of changes and I want them to stabilize first.

CJ-Wright:
Has there been any progress on this? Is there something we can to do help?

asmeurer:
I haven't had the time to work on it. I'm pretty sure it's ready to go, but I just want to do some testing before I merge, as I remember the changes here being somewhat finicky.

asmeurer:
OK, it seems to work. I'm merging. Will try to get a release out some time this week (hopefully tomorrow).

CJ-Wright:
WOOO thank you so very much!

PR #239, opened 2017-08-22
Print an error message when doctr fails 

asmeurer:
Some commands are run in the finally block, which makes it hard to see where the actual error is. This hopefully makes it easier to find in the log.

I have a test error in now to test it out. When I'm satisfied I'll revert it.


asmeurer:
I wish Travis would highlight stderr somehow. I guess I can just print the ANSI color codes, since we know it will be run in Travis.

asmeurer:
And the actual error message is still printed away from the ERROR line.

asmeurer:
I think this is a good as I can get it, without rewriting quite a bit.

asmeurer:
Travis seems to have glitched out in the error message, but I don't think it's my fault.

PR #241, opened 2017-08-22
Actually sync the drdoctr.github.io build

asmeurer:
This won't work until we add a deploy key, which we can't do until we make it so we can use more than one deploy key.


asmeurer:
It doesn't fail because "Don't have permission to push. Not trying."

So we should probably be a little more aggressive about that.

asmeurer:
This is being handled at https://github.com/drdoctr/doctr/pull/280

PR #246, opened 2017-08-29
Render code in docs properly

CJ-Wright:


asmeurer:
Looks good, but we should figure out https://github.com/drdoctr/doctr/issues/245 before we merge any PRs.

asmeurer:
I didn't realize this was the command line help. This will be wrong for anyone who copies the text from the terminal. It also doesn't fix the real issue, which is that the `-` renders as a bullet on Sphinx.

asmeurer:
Better fix at https://github.com/drdoctr/doctr/pull/249

PR #250, opened 2017-09-01
Don't capture the command output when there is no token

asmeurer:
This changes the API of run_command_hiding_token to only return the
returncode, and always print the output.

This change is done so that the command output in Travis appears in order.
Unfortunately, I don't know how to do this properly when there is a token, but
that's not a common use-case.


asmeurer:
OK, this is better. The commands run from doctr are now printed in blue, which makes them much easier to spot (except when Travis's color detection glitches out). And the command output appears to be in actual order now.

PR #252, opened 2017-09-02
Allow --built-docs to be a file

### LINKED ISSUE #220: --built-docs can't be a file
We should allow just a single file to --built-docs.

asmeurer:
Fixes #220.


asmeurer:
It works.

PR #254, opened 2017-09-08
Update changelog for next release

asmeurer:


asmeurer:
Want to do a release soon. https://github.com/drdoctr/doctr/issues/242 and https://github.com/drdoctr/doctr/pull/225 still need to be done but since there are so many changes I think it's better to wait til the next release.

asmeurer:
OK, 1.6.0 is released. Conda-forge feedstock will be updated as soon as it builds.

PR #259, opened 2017-09-27
Make --command run in the deploy branch again

asmeurer:
This reverts a regression in the 1.6 release, which made --command run from
the main branch. This is less useful, since you can just run a command before
running doctr if you want to run something in the deploy branch. Running in
the deploy branch is what's needed (e.g., SymPy's docs run a script in the
docs repo that modifies the html files before committing).

I changed the test in .travis.yml to run in the branch build, which adds a
file to the branch deploy directory (since adding a file to the index from
--command to deploy it is supposed to a supported way to sync extra files).

See discussion at https://github.com/drdoctr/doctr/pull/192.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,9 @@ script:
       cd ..;
       python -m doctr deploy --sync .;
       python -m doctr deploy --sync --gh-pages-docs docs;
-      python -m doctr deploy --sync --no-require-master  --built-docs docs/_build/html "docs-$TRAVIS_BRANCH";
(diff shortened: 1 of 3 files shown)


asmeurer:
Going to do a bug release if/when this passes, since this is a regression, and it's breaking the SymPy builds (possibly others; I don't know who else is using --command).

asmeurer:
OK, it works.

asmeurer:
I'll just do the release off this branch.

PR #265, opened 2017-10-17
Use exit=False in the retry logic

asmeurer:
Otherwise, it won't actually retry if the commands fail, it will just kill
doctr.


asmeurer:
The latest master build definitely confirms that this isn't currently working.

PR #266, opened 2017-10-18
Fix diff-index check

asmeurer:
For whatever reason this fails sometimes. 

Fixes https://github.com/drdoctr/doctr/issues/262

Here is the working build https://travis-ci.org/sympy/sympy/jobs/289271045

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -455,7 +455,7 @@ The doctr command that was run is
     )
 
     # Only commit if there were changes
-    if run(['git', 'diff-index', '--exit-code', '--quiet', 'HEAD', '--'], exit=False) != 0:
+    if run(['git', 'diff-index', '--exit-code', '--cached', '--quiet', 'HEAD', '--'], exit=False) != 0:
         print("Committing")
         run(['git', 'commit', '-am', commit_message])
         return True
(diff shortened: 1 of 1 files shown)


asmeurer:
The --exit-code should be redundant with --quiet. --cached makes it only look at the index. I guess for SymPy, the generate_indexes.py command was changing files but not modifying them. The documentation says that the command should add any changes to be committed to the index, so this is consistent with that. The doctr code is assuming that changes are added, so it makes sense that this check should check just the index, not all uncommitted changes.

asmeurer:
I'm a little unsure this logic is correct, though it does seem to fix the issue I'm having (but I also don't really see how). We may need to revisit the way --command works again in the future, particularly with how it relates to the git index.

asmeurer:
OK, so the way the code works, it only actually commits changes that are in the index after the diff-index check. So using `--cached` is technically correct.

Now, should doctr be smarter about detecting changes after --command, in case they aren't added properly? Probably so. Requiring the command to add all files, not just new ones, to the index, is cumbersome and error prone (it is *only* necessary if the file changed isn't already changed by the docs sync). 

I believe this fixes my immediate issue, and as far as I know, SymPy is the only repo using --command (could be wrong about that, though). It shouldn't be a regression, though, because if the command is currently changing files that aren't added, it still doesn't commit them (and if no files actually are added, it errors). 

I opened https://github.com/drdoctr/doctr/issues/267 to track this.

PR #276, opened 2017-11-10
Add support for multiple deploy key and GitHub Wiki

### LINKED ISSUE #242: Multiple deploy keys
We currently cannot support multiple deploy keys. You need this, for instance, if you want to push to gh-pages for drdoctr/doctr and also to master of drdoctr/drdoctr.github.io (which we do want to do, to test it). See https://github.com/drdoctr/doctr/pull/190.

You can easily have two different deploy private key files, but the problem is that we encrypt that with a secret encryption key, which is stored in DOCTR_DEPLOY_ENCRYPTION_KEY. That key is generated when doctr configure is run, and is encrypted with the Travis public key. 

So we either need to:

1. Allow multiple encryption keys (like rename the environment variable based on the repo it is encrypted for).
2. Allow doctr configure to create multiple deploy keys at once, each encrypted with the same encryption key.

### LINKED ISSUE #275: Support GitHub wikis
GitHub wikis are backed by a git repository (e.g., https://github.com/drdoctr/doctr.wiki.git). I haven't checked, but assuming a deploy key allows pushing to a wiki, we should add support for it. This would likely just be detecting if the repo name ends in `.wiki` and stripping it when uploading the key in doctr configure.

ylemkimon:
* Add support for multiple deploy key encryption key. Automatically determines deploy key encryption key environment variable name and filename from deploy key repo name: Fixes #242.
* Add support for deployment to GitHub wiki: Fixes #275.

### DIFF
--- a/doctr/local.py
+++ b/doctr/local.py
@@ -245,6 +245,11 @@ def check_repo_exists(deploy_repo, service='github', *, auth=None, headers=None)
     else:
         raise RuntimeError('Invalid service specified for repo check (neither "travis" nor "github")')
 
+    wiki = False
+    if repo.endswith('.wiki') and service == 'github':
+        wiki = True
+        repo = repo[:-5]
+
     r = requests.get(REPO_URL.format(user=user, repo=repo), auth=auth, headers=headers)
 
     if r.status_code == requests.codes.not_found:
@@ -253,8 +258,17 @@ def check_repo_exists(deploy_repo, service='github', *, auth=None, headers=None)
                                                                            service=service))
 
     r.raise_for_status()
+    private = r.json().get('private', False)
+    
+    if wiki and not private:
+        # private wiki needs authentication, so skip check for existence
+        p = subprocess.run(['git', 'ls-remote', '-h', 'https://github.com/{user}/{repo}.wiki'.format(
+            user=user, repo=repo)], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
+        if p.stderr or p.returncode:
+            raise RuntimeError('Wiki not found. Please create a wiki')
+        return False
 
-    return r.json().get('private', False)
+    return private
 
 GIT_URL = re.compile(r'(?:git@|https://|git://)github\.com[:/](.*?)(?:\.git)?')
 
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -139,7 +139,7 @@ options available.
         (right now only works for Sphinx docs).""")
     deploy_parser.add_argument('--deploy-branch-name', default=None,
                                help="""Name of the branch to deploy to (default: 'master' for ``*.github.io``
-                               repos, 'gh-pages' otherwise)""")
+                               and wiki repos, 'gh-pages' otherwise)""")
     deploy_parser_add_argument('--tmp-dir', default=None,
         help=argparse.SUPPRESS)
     deploy_parser_add_argument('--deploy-repo', default=None, help="""Repo to
@@ -250,7 +250,7 @@ def deploy(args, parser):
     if args.deploy_branch_name:
         deploy_branch = args.deploy_branch_name
     else:
-        deploy_branch = 'master' if deploy_repo.endswith(('.github.io', '.github.com')) else 'gh-pages'
+        deploy_branch = 'master' if deploy_repo.endswith(('.github.io', '.github.com', '.wiki')) else 'gh-pages'
 
     current_commit = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('utf-8').strip()
     try:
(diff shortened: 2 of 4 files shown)


asmeurer:
Thanks. I'll try to take a look soon. Is the multiple deploy key thing required for the wiki?

asmeurer:
Oh right I guess it would be required for SymPy, since we already use doctr for the docs, which are in a separate repo.

ylemkimon:
@asmeurer Or, now I think about it, we can put it to the other build than Sphinx build(`TEST_SPHINX`) in SymPy.

asmeurer:
Indeed. Do per-matrix variables already work in the existing doctr? I don't think I've tested this.

ylemkimon:
@asmeurer In matrix-include-env, it works fine. However, in env-matrix, it has to be encrypted with the plain variables like `travis-encrypt SUPER_SECRET="SECRET";SPLIT="1/4";TEST_SYMPY="true"` and Travis will only recognize SUPER_SECRET as environment variable, but as it is processed as bash statement, other environment variables will be set.

asmeurer:
Wait it really lets you encrypt more than one environment variable just by putting semicolons in them?

ylemkimon:
Yes, also I think spaces also will work if we directly encrypt it. However, since it is encrypted, it will not be available to PR builds.

ylemkimon:
@asmeurer Yes, I think spaces will also work if we directly encrypt it. Also Travis doesn't recognize it as secure variable, so it does not mask it. However, since it is encrypted, it will not be available to PR builds.

doctr/local.py
```diff
@@ -253,8 +258,17 @@ def check_repo_exists(deploy_repo, service='github', *, auth=None, headers=None)
                                                                                     service=service))
 
     r.raise_for_status()
+    private = r.json().get('private', False)
+    
+    if wiki and not private:
+        # private wiki needs authentication, so skip check for existence
+        p = subprocess.run(['git', 'ls-remote', '-h', 'https://github.com/{user}/{repo}.wiki'.format(
+            user=user, repo=repo)], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
```
ylemkimon: Maybe we can do:
```python
p = subprocess.run(['git', 'ls-remote', '-h', 'https://{auth}@github.com/{user}/{repo}.wiki'.format(
    auth=base64.urlsafe_b64encode(bytes(auth.username + ':' + auth.password, 'utf8')).decode(),
    user=user, repo=repo)], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=False)
```

asmeurer:
OK, I did some refreshing on #242. I think if we are going to do refactoring there, we should go full option 3 from that issue, that is, storing the full SSH key in the environment variable (automatically named `DOCTR_DEPLOY_KEY_ORG_REPO` or something like that). I don't know if we'll need the key flag that's been added here after that. I can try to do some work on this this week, unless you want to work on it. 

By the way, the semicolon thing isn't useful because we have to encrypt the keys separately (they are generated by different `doctr configure` runs in general). The matrix thing would be nice, but I couldn't figure out how to make it work (see https://stackoverflow.com/questions/45993334/multiple-values-of-a-secure-encrypted-environment-variable-in-travis-yml). At any rate, that change wouldn't really be relevant to anything other than the documentation and help output of `doctr configure`. The variable itself would still be set just the same to `doctr deploy`. The main difference is that it would be more secure to have the secure variable only available for the specific builds that need it.

ylemkimon:
> I can try to do some work on this this week, unless you want to work on it.
If you don't mind I'd like to work on this.

> The matrix thing would be nice, but I couldn't figure out how to make it work
We can include it in the matrix like (excerpt from sympy/sympy):
```
matrix:
  include:
    - python: 3.6
      env:
        - TEST_ASCII="true"
        - TEST_OPT_DEPENDENCY="numpy scipy gmpy2 matplotlib theano llvmlite autowrap python-symengine=0.3.* tensorflow numexpr ipython"
        - secure: [encrypted 1]

    - python: 3.6
      env:
        - TEST_SPHINX="true"
        - FASTCACHE="false"
        - secure: [encrypted 2]
```
I think this probably work in most use cases, since usually deployment build of documentation is separate build.

ylemkimon:
@asmeurer I'm afraid that Travis is using RSA-2048 PKCS #1 for encrypting variables and it can encrypt maximum 245 bytes([RFC 2313](https://www.ietf.org/rfc/rfc2313.txt)). So we cannot remove Fernet encryption and it will be just including encrypted deploy key(`.enc` file) in `travis.yml` which, I think, is bad for readability of `travis.yml`.

asmeurer:
I see. So that explains why it's always been done this way. 

So we should just use `DOCTR_DEPLOY_ENCRYPTION_KEY_ORG_REPO`, and the default file name should probably encorporate the org/repo as well. And we need to keep backwards compatibility with the current scheme.

ylemkimon:
@asmeurer Added new commits. It automatically determine deploy key encryption key environment variable name and filename from deploy key repo name.

asmeurer:
OK, I finally have time to work on this. 

In order to test this, I've needed to push it up to the main repo. I made a new PR https://github.com/drdoctr/doctr/pull/280. I'm closing this for now. If you want to make more changes you can PR them to the `wiki` branch.

ylemkimon:
@asmeurer It seems it is possible to set long secure environment variables in Travis repository setting, but I haven't tested it.

asmeurer:
Is there an API for it?

ylemkimon:
@asmeurer Yes and it seems that #176 implements it.

ylemkimon:
The downside is that it is not version controlled and it can be only accessed by the repo's owner.

asmeurer:
So I guess the current method is worth keeping? 

@gforsyth what do you think?

asmeurer:
> can be only accessed by the repo's owner.

Like you need GitHub owner permissions, not just push access? 

Currently to add a deploy key to a GitHub repo you need push access (I believe, or does it require full admin?).

ylemkimon:
@asmeurer It needs full admin(owner) access to add a deploy key, too, sorry. I was thinking more like, we or users need another layer of encryption to configure with `--no-upload-key` and [setup Doctr for a repo you don’t have admin access to](https://drdoctr.github.io/doctr/recipes.html#setting-up-doctr-for-a-repo-you-don-t-have-admin-access-to), or owner has to `doctr configure` him/herself.

PR #284, opened 2017-11-20
Add a global table of contents to the docs sidebar

asmeurer:
This fixes https://github.com/drdoctr/doctr/issues/231. If we find a better theme, we can use it, but this is the main thing I wanted.

### DIFF
--- a/docs/_templates/sidebarhelp.html
+++ b/docs/_templates/sidebarhelp.html
@@ -1,0 +1,6 @@
+<h3>Need help?</h3>
+
+<p>
+  Open an issue in our <a href="https://github.com/drdoctr/doctr/issues">issue
+  tracker</a>. Issues that are just questions are fine.
+</p>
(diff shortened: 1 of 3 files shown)


ylemkimon:
It seems local toc and global toc have the same name.

asmeurer:
I'm not following. 

I added both to `index` becaause for some reason the global toc doesn't include the local toc there (it works on every other page).

ylemkimon:
toc

asmeurer:
I see what you mean. I'll have to see if that's configurable.

asmeurer:
It seems there is not a good way to include `index` itself in the global toc (see https://github.com/sphinx-doc/sphinx/issues/2103). But I think we can achieve the effect by copying the templates for the tocs and modifying them.

asmeurer:
I think I found a reasonable solution at https://github.com/drdoctr/doctr/pull/285 (see https://drdoctr.github.io/doctr/docs-release/). It isn't perfect but it's better (see the commit message of https://github.com/drdoctr/doctr/pull/285/commits/c8096abb12ea718eb9deb2b6416c5c9df652d355).

PR #296, opened 2018-02-01
Add --exclude flag to doctr deploy

asmeurer:
This isn't ready yet (I need to add some tests).


asmeurer:
I'm running a live test at https://travis-ci.org/sympy/sympy.github.com/builds/337885302. If it works properly, I'll merge this and do a bugfix release.

asmeurer:
Ah, it doesn't do the right thing with directories.

asmeurer:
Is there an easy way to check if a path is contained in a list of other paths (including being a subdirectory)? Apparently os.path doesn't even have a subdirectory checking function.

asmeurer:
OK, it worked for the SymPy site https://travis-ci.org/sympy/sympy.github.com/builds/338327612. 

As a final test, when merged, https://drdoctr.github.io/doctr/exclude/index.html should exist, but https://drdoctr.github.io/doctr/exclude/test.html should not.

PR #298, opened 2018-02-06
Fix Travis API request

asmeurer:
For whatever reason, the old method startted giving 404 response codes. We now
use API version 3, which has a new URL format, and requires that the org/repo
be entirely url encoded (including the slash).

### DIFF
--- a/doctr/local.py
+++ b/doctr/local.py
@@ -9,6 +9,7 @@ import base64
 import subprocess
 import re
 from getpass import getpass
+import urllib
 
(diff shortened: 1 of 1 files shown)


asmeurer:
@mtazzari do you want to test?

asmeurer:
Looks like our tests actually catch this (in master they fail). So I guess this was broken very recently.

mtazzari:
Just tested it, works perfectly.

PR #302, opened 2018-04-10
Generate the ssh key using the cryptography module

### LINKED ISSUE #301: Doctr configure command doesn't work locally on Windows due to lack of native ssh-keygen executable
I'm a contributor with the [Spyder IDE](https://github.com/spyder-ide), and we're currently completing the process of transitioning our docs to be deployed using ``doctr`` (I'll send a PR to add our name as soon as they officially are declared production-status very soon). When running ``doctr configure`` on Windows, key generation failed due to it being unable to find a ``ssh-keygen`` binary on the PATH. Therefore, it seems ``doctr`` presumes it is being run on a *nix system with this available, even though this is not indicated anywhere I could find in the documentation and unfortunately, a significant portion of developers are stuck on Windows for one reason or another. 

After figuring out the root cause of the issue, I was able to hack a quick fix by adding the appropriate directory from my ``msysgit`` install to the $PATH$ first which had it (presumably my Cygwin install would have worked too, or anything else that had that binary). However, this obviously is pretty poor Ux for a new user, if this is in fact a problem for any Windows user that doesn't have some sort of ``msys`` install or the like on their path.

Therefore, at minimum, the documentation should presumably be updated to reflect this and the fix, or perhaps a more sustainable solution could be devised (e.g. autodetecting an existing ``MSYS`` install like with GfW or Cygwin, etc. or perhaps even WSL, and if not notifying the user interactively of the problem and how to fix it. Thanks!

asmeurer:
This is better because ssh-keygen is not available on all platforms (such as
Windows).

This changes the Python API a bit:

- generate_ssh_key now returns the private and public ssh keys, and does not
  write to a file.

- encrypt_file has been removed and replaced with encrypt_to_file, which
  encrypts contents and writes them to a file.

Fixes #301.

### DIFF
--- a/doctr/__init__.py
+++ b/doctr/__init__.py
@@ -1,4 +1,4 @@
-from .local import (encrypt_variable, encrypt_file, GitHub_post,
+from .local import (encrypt_variable, encrypt_to_file, GitHub_post,
     generate_GitHub_token, upload_GitHub_deploy_key, generate_ssh_key,
     check_repo_exists, guess_github_repo)
 from .travis import (decrypt_file, setup_deploy_key, get_token, run,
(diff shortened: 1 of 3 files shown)


asmeurer:
Ah forgot that PRs need to go from the main repo.

PR #306, opened 2018-06-02
Better logic for falling back to the legacy deploy key filename

asmeurer:
This way if the file isn't found, the new file name is in the error message
rather than the legacy one.


gforsyth:
Hey @asmeurer -- this looks good to me.  Do you want to rebase against master?

PR #310, opened 2018-08-06
Add support for public repos on travis-ci.com

asmeurer:
Fixes https://github.com/drdoctr/doctr/issues/309

This is still not fully tested, and I have a few things that need to be fixed (like the url in the commit message). 

I'm unsure if the logic here is correct. <strike>I test travis-ci.org first and then travis-ci.com. The reason is that travis-ci.com always requires authentication, so .org is preferred. However, maybe it should be the other way around, as apparently it's possible to enable both. </strike>

### DIFF
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -248,7 +248,7 @@ def process_args(parser):
     try:
         return args.func(args, parser)
     except RuntimeError as e:
-        sys.exit("Error: " + e.args[0])
+        sys.exit(red("Error: " + e.args[0]))
 
 def on_travis():
     return os.environ.get("TRAVIS_JOB_NUMBER", '')
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -196,9 +196,14 @@ options available.
         public repositories for the user. This option is not recommended
         unless you are using a separate GitHub user for deploying.""")
     configure_parser.add_argument("--no-upload-key", action="store_false", default=True,
-        dest="upload_key", help="""Don't automatically upload the deploy key to GitHub. If you select this
-        option, you will not be prompted for your GitHub credentials, so this option is not compatible with
-        private repositories.""")
+        dest="upload_key", help="""Don't automatically upload the deploy key to GitHub. To prevent doctr
+        configure from asking for your GitHub credentials, use
+        --no-authenticate.""")
+    configure_parser.add_argument("--no-authenticate", action="store_false",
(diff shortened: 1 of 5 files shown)


asmeurer:
I think I need to fix the logic to check if both .com and .org exist and ask in that case. Otherwise we have situations like doctr where the .com "exists" but isn't working https://travis-ci.com/drdoctr/doctr, or conversely, repos like the one in the issue, or drdoctr/travis-ci-com-testing that exist and are run on both (https://travis-ci.com/drdoctr/travis-ci-com-testing, https://travis-ci.org/drdoctr/travis-ci-com-testing). I don't think we need to actually support both, as assumedly this situation won't last forever.

asmeurer:
It doesn't seem easy to detect whether or not we are running in .com or .org for the commit message (https://github.com/travis-ci/travis-ci/issues/8935, https://github.com/travis-ci/travis-ci/issues/7552). I'll play with it some more and see if it's possible to extract it.

asmeurer:
I'm not seeing a way, other than saving it somehow in the configure stage, which is annoying considering it's just for the commit message.

asmeurer:
I guess this is ready for review. I still need to do a lot more testing of it. 

Unfortunately, Travis never got back to me on my questions, so for now:

- Configuring travis-ci.com will require authentication. 
- The URLs in the commit messages for travis-ci.com builds will be wrong (this has already the case for private repo builds on .com ever since they have been supported). You can fix them by replacing `.org` with `.com`.

asmeurer:
I think this is ready to go. The travis-ci.com testing is at https://travis-ci.com/drdoctr/travis-ci-com-testing. 

I've also split out a new `--no-authenticate` flag for `doctr configure`, which implies `--no-upload`. `--no-upload` no longer doesn't ask for authentication without `--no-authenticate`. If `--no-authenticate` is used with a private repo or with a travis-ci.com repo, it is an error. Hopefully Travis can re-enable the ability to get the public key without authenticating, as it isn't a particularly user friendly process. Not only does it make it impossible for someone to add doctr to a repo they don't have access to, it requires temporarily creating a personal access token, which results in a notification email from GitHub (we delete it immediately, but it's still confusing to users). 

I'll let this sit for a bit in case anyone wants to review/test. Otherwise I'll merge and get a release out, probably tomorrow or Thursday. I'm still unsure but likely https://github.com/drdoctr/doctr/pull/332 should also go in the release.

asmeurer:
Oh I forgot to add some flags to specify which to use from the command line.

Also it looks like there is a bug for repos that aren't activated on either yet (KeyError).

PR #315, opened 2018-08-17
Added a recipe to explain separating source and output in user pages.

moorepants:
It is not obvious how to separate the source from the output files in
two branches when using Github username or orgname root pages. This
additional recipe explains how to do this.

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -183,22 +183,43 @@ The deploy key for pushing to a wiki is the same as for pushing to the repo
 itself, so if you are pushing to both, you will not need more than one deploy
 key.
 
-Separating source and output in Github username and orgname pages
-=================================================================
+Using doctr with Github username and orgname pages
+==================================================
 
-Github allows users to create pages at the root url of users or organizations
-http://github.io pages. For example an organization ``coolteam`` can setup a
-repository with the name ``coolteam.github.io`` and the html files in the
-master branch of this repository will be served to https://coolteam.github.io.
-If you are using a static site generator you may want to separate your web
-site's source files from the generated output. To do this, create a new branch,
-e.g. ``source``, and push this to Github. Set this branch as the default branch
-in the Github settings for the repository. Only commit source files to this
-branch and run the ``doctr configure`` command in this branch committing the
-generated key and the necessary ``.travis.yml`` file. The source and output
-repositories should be set to ``coolteam/coolteam.github.io`` in the
-configuration questions. Lastly, in the ``.travis.yml`` make sure that the
-``doctr deploy`` command has the flag: ``--branch-whitelist source`` if
-``source`` is the default branch name. Change ``source`` to whatever your
-default branch is set to. All output files will be pushed to the master branch
-during the Travis builds and the source is separated from the output.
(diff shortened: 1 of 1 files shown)


docs/recipes.rst
```diff
@@ -182,3 +182,23 @@ as the deploy repo running ``doctr configure``. When deploying, use
...
+
+Separating source and output in Github username and orgname pages
+=================================================================
+
+Github allows users to create pages at the root url of users or organizations
+http://github.io pages. For example an organization ``coolteam`` can setup a
+repository with the name ``coolteam.github.io`` and the html files in the
+master branch of this repository will be served to https://coolteam.github.io.
+If you are using a static site generator you may want to separate your web
+site's source files from the generated output. To do this, create a new branch,
```
asmeurer: Actually as you saw, currently with doctr you have to do this. Either put the sources on a separate branch or on a separate repo (there's already [a recipe](https://drdoctr.github.io/recipes.html#deploy-to-a-separate-repo) for separate repo).

docs/recipes.rst
```diff
@@ -182,3 +182,23 @@ as the deploy repo running ``doctr configure``. When deploying, use
...
+e.g. ``source``, and push this to Github. Set this branch as the default branch
+in the Github settings for the repository. Only commit source files to this
+branch and run the ``doctr configure`` command in this branch committing the
+generated key and the necessary ``.travis.yml`` file. The source and output
+repositories should be set to ``coolteam/coolteam.github.io`` in the
+configuration questions. Lastly, in the ``.travis.yml`` make sure that the
+``doctr deploy`` command has the flag: ``--branch-whitelist source`` if
```
asmeurer: Add this as a code block, like the other recipes.

asmeurer:
There's at least two places that doctr itself could be smarter here. I've opened issues for them https://github.com/drdoctr/doctr/issues/316 https://github.com/drdoctr/doctr/issues/317

docs/recipes.rst
moorepants: You don't have too. I first had the sources and output files mixed all in the master branch. It works but you can possibly get clashes if you aren't careful with file and directory names, etc.

moorepants: Oh yeah, but yes the infinite loop was probably going in the sources + outputs in master branch. So you shouldn't do that.

moorepants:
Thanks. I'll edit and submit again.

docs/recipes.rst
asmeurer: With https://github.com/drdoctr/doctr/pull/318 you will be able to, but even then it's not ideal because every commit will result in a redundant build.

asmeurer: Plus if you put your source files on `master` GitHub will serve all your source files on GitHub pages. That probably isn't a big deal but it could be unexpected.

moorepants: Yeah. I changed the wording to point the user away from doing this.

asmeurer:
Great. I like the updated text. We can remove the warning once https://github.com/drdoctr/doctr/pull/318 is merged.

moorepants:
Thanks! Great tool :)

Issue #316, opened 2018-08-17
Don't run doctr on a commit generated by doctr


asmeurer:
See https://github.com/drdoctr/doctr/issues/314. If you accidentally run Travis on the GitHub pages branch, then doctr will run into an infinite loop, triggering a build each time it pushes a commit. This is generally a misconfiguration, but we should avoid running doctr when this happens. Maybe it should be an error by default to alert the user that they messed up. We can check that the commit email is the one that doctr sets (doctr@noreply.github.com or something like that).

asmeurer:
Fixed by https://github.com/drdoctr/doctr/pull/318

PR #318, opened 2018-08-17
Check if `doctr` committed `HEAD` before trying to push

gforsyth:
In reference to #316

This only prints a warning if the last commit was by `doctr` but it could be an error instead.  Thoughts?

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -256,14 +256,14 @@ def set_git_user_email():
     """
     Set global user and email for git user if not already present on system
     """
-    username = subprocess.run(shlex.split('git config user.name'), stdout=subprocess.PIPE).stdout.strip()
-    if not username or username == b"Travis CI User":
+    username = subprocess.run(shlex.split('git config user.name'), stdout=subprocess.PIPE).stdout.strip().decode('utf-8')
+    if not username or username == "Travis CI User":
         run(['git', 'config', '--global', 'user.name', "Doctr (Travis CI)"])
     else:
         print("Not setting git user name, as it's already set to %r" % username)
 
-    email = subprocess.run(shlex.split('git config user.email'), stdout=subprocess.PIPE).stdout.strip()
-    if not email or email == b"travis@example.org":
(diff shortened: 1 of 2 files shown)


doctr/travis.py
```diff
@@ -534,6 +534,15 @@ def push_docs(deploy_branch='gh-pages', retries=3):
             return
     sys.exit("Giving up...")
 
+def last_commit_by_doctr():
+    """Check whether the author of `HEAD` is `doctr` to avoid starting an
+    infinite loop"""
+
+    email = run(["git", "show", "--format=%ae", "HEAD"])
```
asmeurer: `run` returns the returncode. This needs to be `subprocess.check_output`.

asmeurer:
I forgot but Travis already doesn't run on branches named `gh-pages`.

asmeurer:
Looks like this doesn't work because we only change the user if it isn't set, and I guess Travis sets the user now.

asmeurer:
OK it works https://travis-ci.org/drdoctr/drdoctr.github.io/builds/417488916#L859. We should change the doctr tagline to "we bang our heads against the wall trying to get Travis to work so you don't have to".

Good thing I decided to actually test this. I'm going to leave up the tests at https://travis-ci.org/drdoctr/drdoctr.github.io/builds for the time being. It shouldn't mess up our Travis queue too much because of automatic build cancellation, meaning it basically won't run until the doctr queue is cleared as long as those builds keep pushing to it. If it ends up building infinitely in the future it means this got broken somehow. 

In unrelated news, I finally spotted a successful push retry in the wild https://travis-ci.org/drdoctr/doctr/jobs/417474729#L2661

asmeurer:
I also did some minor updates to @moorepants's recipe here. Will merge once the builds finish.

asmeurer:
Apparently our username setting code has been broken for a while. The first commit on sympy_doc that used the Travis CI user was from September of last year https://github.com/sympy/sympy_doc/commit/70d60cc7cac0daa468362152a7b1eec9ef704dfc

PR #320, opened 2018-08-27
Better messaging on 401 and 403 errors from GitHub

### LINKED ISSUE #311: 403 client error on api.github.com
I am trying to set up doctr on a new repo following the instructions in the readme. I have the latest version of doctr (1.7.3).

After I enter my authentication code, I get a 403 Client Error.

```
$ doctr configure
Welcome to Doctr.

We need to ask you a few questions to get you on your way to automatically
deploying from Travis CI to GitHub pages.

What is your GitHub username? rabernat
Enter the GitHub password for rabernat: 
A two-factor authentication code is required: sms
Authentication code: ******
Traceback (most recent call last):
  File "/Users/rpa/miniconda3/envs/geo_scipy/bin/doctr", line 6, in <module>
    sys.exit(doctr.__main__.main())
  File "/Users/rpa/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/doctr/__main__.py", line 517, in main
    return process_args(get_parser(config=config))
  File "/Users/rpa/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/doctr/__main__.py", line 249, in process_args
    return args.func(args, parser)
  File "/Users/rpa/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/doctr/__main__.py", line 383, in configure
    login_kwargs = GitHub_login()
  File "/Users/rpa/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/doctr/local.py", line 141, in GitHub_login
    return GitHub_login(username=username, password=password, OTP=OTP, headers=headers)
  File "/Users/rpa/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/doctr/local.py", line 145, in GitHub_login
    r.raise_for_status()
  File "/Users/rpa/miniconda3/envs/geo_scipy/lib/python3.6/site-packages/requests/models.py", line 935, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://api.github.com/
```

I am totally stuck on this issue! Any help would be greatly appreciated.

asmeurer:
401 errors are caused when the OTP expires from a long session. The way the
code is currently organized, it isn't easy to re-ask for it, so for now we
just print a message to try again. This generally only happens if the user
does something in the middle of running doctr configure, such as going to
enable Travis on the repo. If everything is already configured, a single
session is generally short enough to use the same OTP code.

403 errors occur when the GitHub API rate limit is hit. This can happen when
unauthenticated requests are used (i.e., --no-upload-key), as the limit is 60
global GitHub API requests per IP per hour. For authenticated requests, the
limit is 5000 requests per hour, but this is shared across all oauth
applications. It seems that the Travis "sync account" button consistently
causes this limit to be hit if you have access to many repos (for instance, if
you are a member of the conda-forge organization). So if a user goes to enable
a repo on Travis, then runs doctr configure, they will hit this error.

doctr configure now prints an error message indicating that the rate limit has
been hit and how long it will be until it resets. Unfortunately, there is not
much else we can do here.

Fixes #311.

### DIFF
--- a/doctr/local.py
+++ b/doctr/local.py
@@ -176,7 +176,7 @@ def GitHub_raise_for_status(r):
         def plural(n):
             return n, abs(n) != 1 and "s" or ""
 
-        s = "%02d minute%s" % plural(mm)
+        s = "%d minute%s" % plural(mm)
         if hh:
             s = "%d hour%s, " % plural(hh) + s
         if relative_reset_datetime.days:
--- a/doctr/local.py
+++ b/doctr/local.py
@@ -10,7 +10,6 @@ import re
 from getpass import getpass
 import urllib
 import datetime
-import textwrap
 
 import requests
 from requests.auth import HTTPBasicAuth
--- a/doctr/local.py
+++ b/doctr/local.py
@@ -9,6 +9,8 @@ import subprocess
 import re
 from getpass import getpass
 import urllib
+import datetime
+import textwrap
 
 import requests
 from requests.auth import HTTPBasicAuth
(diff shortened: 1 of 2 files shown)


asmeurer:
Note for anyone testing this, the messages will actually print in red once this and https://github.com/drdoctr/doctr/pull/310 are merged together.

asmeurer:
We can also activate repositories on Travis via the API https://developer.travis-ci.com/resource/repository#activate. We should probably offer to do this if a repo exists but isn't activated on Travis yet.

doctr/local.py
```diff
@@ -130,22 +131,78 @@ def GitHub_login(*, username=None, password=None, OTP=None, headers=None):
...
+        else:
+            message += """
+You can get a higher API limit by authenticating. Try running doctr configure
+again without the --no-upload-key flag.
+"""
+        message += """
+Your rate limits will reset in {s}.\
+""".format(s=s)
+        raise GitHubError(message)
+    r.raise_for_status()
```
moorepants: This all seems clear to me. Should be very helpful.

asmeurer:
I was playing around with doing that, and I hit the API limit again. So I'm not sure but it looks like Travis might hit the GitHub API limits even if you try to activate a repo manually via the API.

PR #332, opened 2018-10-18
Don't attempt to push on builds from forks

### LINKED ISSUE #110: Detect if the repo is a fork
If I'm understanding the behavior I'm seeing correctly, encrypted environment variables aren't available on forks at all, even if they are encrypted with the key for the fork.

### LINKED ISSUE #236: DOCTR_DEPLOY_ENCRYPTIOn_KEY not set
So I tried to follow all the instructions, but got this error
https://travis-ci.org/xpdAcq/redsky#L934
Here is the [`travis.yml`](https://github.com/xpdAcq/redsky/blob/master/.travis.yml)
Any thoughts on what I am doing wrong?
(I do have the [`github_deploy_key.enc`](https://github.com/xpdAcq/redsky/blob/master/github_deploy_key.enc) that doctr made)

### LINKED ISSUE #331: Syncing a fork gives DOCTR_DEPLOY_ENCRYPTION_KEY error
Hello, this is my first time raising an issue, so apologies if I missed something from the API that answers this.

Essentially, I have forked a repo that uses doctr to deploy their documentation to github-pages via Travis. The upstream repo is https://github.com/PAHdb/pyPAHdb, and my fork is https://github.com/mattjshannon/pyPAHdb.

When I sync my fork (master) to the upstream (master) after an accepted PR, Travis attempts a build and then fails with the following output:

(from https://travis-ci.org/mattjshannon/pyPAHdb/builds/443311254)
```
$ doctr deploy --built-docs docs/build/html .
Setting git attributes
git config --global user.name 'Doctr (Travis CI)'
git config --global user.email drdoctr@users.noreply.github.com
Adding doctr remote
ERROR: The doctr command '/home/travis/virtualenv/python3.6.3/bin/doctr deploy --built-docs docs/build/html .' failed: RuntimeError('DOCTR_DEPLOY_ENCRYPTION_KEY_MATTJSHANNON_PYPAHDB or DOCTR_DEPLOY_ENCRYPTION_KEY environment variable is not set',)
```

When I create branches and make pull requests against the upstream repo the build has no trouble, so I'm not sure why it fails when just syncing the fork to upstream.

Any help would be greatly appreciated. Thank you!

asmeurer:
Travis does not make encrypted environment variables available to forks, so
doctr should not attempt to push from them.

I have to check this with the GitHub API, so hopefully that doesn't cause issues. 

Fixes #110
Closes #331
Closes #236

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ install:
 script:
   - set -e
   - set -x
+  - echo $TRAVIS_SECURE_ENV_VARS
   # Note, the commands below need --sync to actually sync, to override sync: False below. Also, every deploy to the main repo needs to have --key-path deploy_key.enc.
   - |
     if [[ "${DOCS}" == "true" ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,6 @@ install:
 script:
   - set -e
   - set -x
-  - echo $TRAVIS_SECURE_ENV_VARS
   # Note, the commands below need --sync to actually sync, to override sync: False below. Also, every deploy to the main repo needs to have --key-path deploy_key.enc.
   - |
     if [[ "${DOCS}" == "true" ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ script:
     fi
   - if [[ "${TESTS}" == "true" ]]; then
       pyflakes doctr;
-      py.test doctr -v;
+      py.test doctr -v -rs;
     fi
 
 doctr:
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ script:
       python -m doctr deploy --sync --key-path deploy_key.enc stash-test/ --built-docs test;
       # Test syncing a single file
       echo `date` >> test-file
-      python -m doctr deploy --sync --key-path deploy_key.enc test-file --built-docs test-file;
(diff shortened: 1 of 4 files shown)


asmeurer:
The tests are on my fork. https://travis-ci.org/asmeurer/doctr/builds/443414925. It looks like there are some other issues because the doctr tests weren't designed to run on a fork build.

asmeurer:
Looks like the API checks fail pretty regularly because of API limits. I'm not sure how feasible this fix is.

asmeurer:
We do call doctr a lot in our tests (13 tests * 2 Python versions per build), and GitHub's unauthenticated rate limite is 60 requests per hour. I don't know how that ends up being split up on Travis, but assuming that each fork gets its own IP most likely most forks wouldn't hit this issue very often.

asmeurer:
We can print a warning, but I'm not too happy with that, as it could confuse people. Is there really not a way to check this without using the API?

asmeurer:
This also contains a minor fix for syncing a single file when the deploy directory doesn't already exist (it came up because my fork has an old copy of `gh-pages`).

asmeurer:
Going to try this. If the API call fails, it assumes that it isn't on a fork, so it may still fail on forks sometimes (or print a warning message on non-forks). 

I've also got some fixes to single-file syncs which came up on my fork https://travis-ci.org/asmeurer/doctr/branches. If we do end up reverting this we should try to keep those fixes intact.

PR #337, opened 2019-01-30
Fix travis-ci.com stuff

### LINKED ISSUE #335: Issues with travis-ci.com
There are some issues with travis-ci.com that weren't fixed at https://github.com/drdoctr/doctr/pull/310

It looks like repos that aren't activated on either lead to a KeyError.

Also, apparently the travis-ci.com API has a key to indicate if a repo is also on .org, but it is currently broken. See https://travis-ci.community/t/active-on-org-from-api-v3-is-wrong/1199.

asmeurer:
It apparently doesn't require authentication to use travis-ci.com, unless the repo is private, so this simplifies things a lot. 

I am still testing this out. I should figure out how to create a private repo to test that as well. I had thought GitHub had free private repos now, but it won't let me create one on drdoctr. 

Fixes #335.

### DIFF
--- a/docs/tests.rst
+++ b/docs/tests.rst
@@ -21,3 +21,49 @@ You can add the updated token to Travis on the `doctr Travis Settings Page
 Paste the token string into the ``Value`` field and ``TESTING_TOKEN`` in the
 ``Name`` field (unless you have changed this value in
 ``doctr/tests/test_local.py``).
+
+travis-ci.com cs. travis-ci.org
+-------------------------------
+
+Travis CI is `migrating
+<https://blog.travis-ci.com/2018-05-02-open-source-projects-on-travis-ci-com-with-github-apps>`_
+from .org to .com. While the migration is in place, it is possible to enable a
+repository on both. However, the same repository on each will have different
+public keys. Doctr presently only supports one at a time. On Travis itself,
+there is little difference in the doctr code, but there is a lot of code in
+``configure`` to automatically determine which is enabled.
+
+The repos:
+
+- https://github.com/drdoctr/testing-travis-ci-com
+- https://github.com/drdoctr/testing-travis-ci-org
+- https://github.com/drdoctr/testing-travis-ci-both
+- https://github.com/drdoctr/testing-travis-ci-neither
+
+Are enabled on Travis CI .com, .org, both, and neither. To enable a repo on
+.org, go to https://travis-ci.org/organizations/drdoctr/repositories and make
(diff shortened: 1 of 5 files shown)


asmeurer:
There are now four drdoctr/testing-travis-ci-* repos to test the four combinations. Each repo indicates how it should be set up in its README. These are used in the test suite to make sure `check_repo_exists` does the right thing.

asmeurer:
I guess I can create one, but only on my personal account. It won't let me create it on drdoctr. 

I've made https://github.com/asmeurer/doctr-private-repo-testing (no one but me will be able to see it). I need to figure out how to get the Travis testing token so it can access it.

asmeurer:
I don't understand why the tests are failing. They work for me locally, after commenting out the skip. Are Travis API requests results different when run on Travis itself?

asmeurer:
Oh the token is for GitHub, not Travis. So it fails because of the token.

asmeurer:
Apparently I only get 100 trial builds for my private repo before I have to pay Travis. So hopefully I get this fixed soon.

asmeurer:
And I guess you have to pay GitHub to use pages on a private repo. But you can still confirm that the push works on the `gh-pages` branch.

asmeurer:
It looks like the private repo build works, both for a token and a deploy key.

asmeurer:
I want to get some build testing on the doctr/testing-* repos, just to make sure the whole flow works correctly. Unfortunately, the way this is designed, only one of travis-ci.com/.org is supported at a time, so we can't automatically test both on this repo. I think it shouldn't be a big deal, as virtually all the complexity there is on the `configure` side. The only difference on Travis is the URL that is included in the commit message.

asmeurer:
OK, I tested it all and it all seems to work. 

Remarkably, you can actually enable a repo on both .com and .org and deploy from both. You just need to add both secure environment variables to .travis.yml, and then rename the encryption key files so that you have both of them, and then do something like

```yaml
  - |
    if [[ $TRAVIS_BUILD_WEB_URL == *"travis-ci.com"* ]]; then
        doctr deploy --built-docs test . --key-path github_deploy_key-com.enc;
    else
        doctr deploy --built-docs test . --key-path github_deploy_key-org.enc;
    fi
```

You can see that it works

https://travis-ci.org/drdoctr/testing-travis-ci-both
https://travis-ci.com/drdoctr/testing-travis-ci-both
https://github.com/drdoctr/testing-travis-ci-both/commits/gh-pages

I can't see a legitimate use-case for this, so I'll probably only mention it in the release notes.

Issue #342, opened 2019-04-05
Getting "Don't have permission to push. Not trying." message


kyleniemeyer:
Although I haven't changed anything except to add some new files, I'm now getting a `Don't have permission to push. Not trying.` message in Travis, which seems to prevent pushing the generated files back to the repo for GitHub Pages to pick up.

Looking at a recent successful build, it looks like that time it did a `git pull` then `git push`, neither of which is happening now.

Any ideas?

asmeurer:
It should print one of these messages indicating why it thinks it shouldn't push: https://github.com/drdoctr/doctr/blob/0f19ff78c8239efcc98d417f36b0a31d9be01ba5/doctr/travis.py#L571-L602

asmeurer:
The exception is if you are using `--no-push` or have `doctr: push: false` in your `.travis.yml`. It won't print a message indicating why in that case.

kyleniemeyer:
OK, I see `The website and docs are not pushed to gh-pages on fork builds`... what does this mean, and why would this be an issue now? (Was this added in a newer version of doctr or something?)

My repo is a fork, but that's because I forked my website template from elsewhere a while ago. How can I override this?

asmeurer:
That's curious if it used to work on the same repo. 

Travis doesn't allow secure environment variables on builds from forked repos (unless things have changed). You may need to delete the repo and recreate it as a non-forked repo.

But again that's curious if it used to work. If that's the case we should figure out what is going on.

kyleniemeyer:
Yeah, it used to work fine, and I haven't changed anything. I hadn't made any changes in a few weeks, that's all.

asmeurer:
OK, I'm guessing the last build was before the most recent release of doctr. The change to not push on forks was new https://drdoctr.github.io/changelog.html, https://github.com/drdoctr/doctr/pull/332. I guess the logic isn't correct, or maybe Travis changed something. I had thought secure environment variables didn't work at all on forks, but I guess that isn't the case. Maybe we should only not push on a fork if the environment variable isn't there.

asmeurer:
If you need a workaround you can manually install doctr 1.7.4.

asmeurer:
Fix started at https://github.com/drdoctr/doctr/pull/343. I still need to fix the tests and do some manual testing. If you can check if that branch works on your repo that would help.

kyleniemeyer:
Can I get Travis to use that one somehow, or should I just try locally?

[...]

PR #347, opened 2019-06-14
TRAVIS_TAG could set canpush without checking all rules

peterjc:
If ``$TRAVIS_TAG`` and ``--build-tags`` are set, then we get ``canpush=True`` without checking all rules, e.g. If the repository was a fork, then the push should be prevented.

As an aside, I noticed this edge can while looking at issue #342 (there are legitimate cases where it is reasonable to want to push from a forked repository).

### DIFF
--- a/doctr/travis.py
+++ b/doctr/travis.py
@@ -575,10 +575,9 @@ def determine_push_rights(*, branch_whitelist, TRAVIS_BRANCH,
     """
     canpush = True
 
-    if TRAVIS_TAG:
-        if not build_tags:
-            print("The docs are not pushed on tag builds. To push on future tag builds, use --build-tags")
-        return build_tags
+    if TRAVIS_TAG and not build_tags:
+        print("The docs are not pushed on tag builds. To push on future tag builds, use --build-tags")
+        canpush = False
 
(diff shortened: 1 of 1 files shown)


asmeurer:
Looks like the doctr tests don't even pass on fork builds. I need to finish https://github.com/drdoctr/doctr/pull/343.

asmeurer:
Oh actually I think that test failure is legitimate. It's a test for this function. So it should be updated.

I think I never considered the fact that someone would push a tag to their fork, but this looks correct.

doctr/travis.py
```diff
@@ -575,10 +575,9 @@ def determine_push_rights(*, branch_whitelist, TRAVIS_BRANCH,
...
 
-    if TRAVIS_TAG:
-        if not build_tags:
-            print("The docs are not pushed on tag builds. To push on future tag builds, use --build-tags")
-        return build_tags
+    if TRAVIS_TAG and not build_tags:
+        print("The docs are not pushed on tag builds. To push on future tag builds, use --build-tags")
+        canpush = False
 
     if not any([re.compile(x).match(TRAVIS_BRANCH) for x in branch_whitelist]):
```
asmeurer: What is TRAVIS_BRANCH set to when TRAVIS_TAG is true?

PR #350, opened 2019-07-14
Put the deploy key directly in the environment variable

asmeurer:
This is https://github.com/drdoctr/doctr/pull/349 but pushed directly to the repo. CC @peterjc


asmeurer:
See https://github.com/drdoctr/doctr/pull/349#issuecomment-511216861

Issue #353, opened 2019-10-18
Add option to transfer post-processing scripts


goerz:
I have a folder that contains a python script that I want to use for post-processing (`doctr deploy --command ...`), along with python modules that the script depends on. Since `doctr` runs the `--command` from within the deploy branch (`gh-pages`), that folder has to be available in doctr's checkout of `gh-pages`. However, I'd very much like to keep the folder on my `master` branch, not `gh-pages`, for a number of reasons:
* Chicken-and-egg problem: I can't set up `doctr` on a new repo that doesn't have a `gh-pages` branch yet: I'd have to create `gh-pages` manually and place the post-processing script in it
* The code for building the docs shouldn't be separated across different branches: the build-commands in my `.travis.yml` and the post-processing script are very much interdependent, so for any changes to the doc-building process I'd have to work simultaneously in two different branches of the project
* A contributor to the project (or myself after a couple of months, having forgotten the details of doctr) might be confused about where to find the post-processing script, and would probably not think to look on the `gh-pages` branch.
* Philosophically, I think of `gh-pages` as a "deployment branch" that I don't want to have to touch by hand.

I would propose to add e.g. a `--transfer` flag to the `doctr` command that would transfer files/folders from the build-branch (`master`/version-tag) to the deploy branch (`gh-pages`). The transferred files/folders should not be commited to `gh-pages`.

For the time being, I'm trying to work around this by copying the folder containing the post-processing script (`.travis`) to the deploy directory (`cp -r .travis docs/build/html/_doctr`) in my `.travis.yml` before the call to `doctr deploy`. However, I'm having a very hard time preventing the `_doctr` directory from being pushed to `gh-pages`. Isn't that what `--exclude _doctr` should do?

asmeurer:
The --exclude thing is probably a bug. 

--transfer as you describe it is already what the deploy directory is. I think we just need to add support for multiple deploy directories/files, and fix --exclude so that it works (and also document this whole thing in the recipes docs).

goerz:
Do you mean multiple `--built-docs` (#221)? So if I had the documentation in `docs/build/html` and my post-processing scripts in `.travis`, I'd invoke `doctr deploy` as something like

~~~
doctr deploy --built-docs docs/build/html --built-docs .travis --exclude .travis $DEPLOY_DIR
~~~

[...]

PR #355, opened 2019-10-20
Double check if files still exist before git-add

goerz:
I figured out what the problem I was having in #353 with deleting files from the deploy directory via a custom --command:  The command runs after syncing, but before committing. Thus, if the command removes any files that were "added" by the syncing, then the "git add" during the commit will fail.

To get around this, we can just double check if "added" files are still there before calling "git add", respectively "removed" files are still deleted. If there are any changes to the status after the sync, a warning seems appropriate.


goerz:
The CI failures do not seem to be related to my patch, cf. the docs-only PR #356 also failing. Let me know if there's anything I should do.

asmeurer:
Yeah, there are some fixes I need to do for the tests. I haven't had the time to work on doctr in awhile unfortunately.

PR #215, opened 2017-07-15
Copy the --built-docs to a temporary directory

asmeurer:
This makes it possible to deploy committed files, and avoids issues with
checkout out gh-pages if tracked files would be modified by checkout.

Fixes #202
Fixes #214
Closes #211

PR #356, opened 2019-10-20
Document the deployment procedure

goerz:
This should clarify some of the details mentioned in the `doctr deploy` documentation, specifically what "synchronization" is (`--no-sync` flag), and how `--command` fits in.


goerz:
Hopefully, I got the procedure right, based on looking over the source code. Let me know if I misunderstood anything.

One question: What is the purpose of the `--no-temp-dir` flag? It skips the first step, copying the `BUILT_DOCS` directory, but I can't think of any situation where this would be useful. I think it would cause a crash unless combined with `--no-sync`. I suppose with `--no-temp-dir` and `--no-sync`, a custom `--command` could take over the entire deployment, but it wouldn't have access to the built documentation (which was stashed away).

asmeurer:
I don't remember why I added it (see #215). It might have been for the reason you noted so that you can control the deployment yourself if you want. But the stash thing was added later, so it might be broken. It looks like the flag isn't tested, so I'm not surprised if it doesn't work.

PR #358, opened 2019-10-26
Fix user and repo tests for Travis

gforsyth:
Not sure of the exact date, but Travis is no longer case-sensitive when
dealing with user and repo names, at least as of the date of this
commit.

Since these are the only tests in `local` we have for checking if a repo
exists in Travis, I've added a random hex-string to the end the repo and
user, respectively, to ensure that the `RuntimeError` is raised as
expected.

### DIFF
--- a/doctr/tests/test_local.py
+++ b/doctr/tests/test_local.py
@@ -1,4 +1,5 @@
 import os
+import binascii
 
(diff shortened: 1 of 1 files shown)


asmeurer:
Oh OK. I thought the test failures had to do with the fork changes, but I guess those are just in the tests for https://github.com/drdoctr/doctr/pull/343.

PR #370, opened 2020-08-02
DOC: warn about --force for dirty repos

stsievert:
**What does this PR implement?**
It warns about running `doctr deploy` with the `--force` flag and dirty files.

I made the mistake of doing this, and this warning would have prevented my mistake. I am submitting this PR to ask a question: how do I recover those dirty files untracked by Git? Or are they gone forever?

### DIFF
--- a/doctr/__main__.py
+++ b/doctr/__main__.py
@@ -271,7 +271,9 @@ def deploy(args, parser):
 
     if not args.force and not on_travis():
         parser.error("doctr does not appear to be running on Travis. Use "
-                     "doctr deploy <target-dir> --force to run anyway.")
+                     "doctr deploy <target-dir> --force to run anyway.\n\n"
+                     "WARNING: this will remove any dirty files"
+                    )
 
     config = get_config()
 


asmeurer:
doctr deploy --force should basically never be used. It forces doctr to run locally, whereas it normally only runs in Travis. It's only useful for development. We should probably hide the flag. 

doctr deploy stashes your dirty changes. It should unstash them, but maybe that failed. You can try running `git stash pop` manually.

stsievert:
> doctr deploy stashes your dirty changes. It should unstash them, but maybe that failed. You can try running git stash pop manually.

Luckily I've been able to recover. `git stash` didn't have my changes, but (very luckily) I've aliased `rm` to move files macOS's "Trash" directory.

doctr/__main__.py
```diff
@@ -271,7 +271,9 @@ def deploy(args, parser):
 
     if not args.force and not on_travis():
         parser.error("doctr does not appear to be running on Travis. Use "
-                     "doctr deploy <target-dir> --force to run anyway.")
+                     "doctr deploy <target-dir> --force to run anyway.\n\n"
+                     "WARNING: this will remove any dirty files"
```
asmeurer, line 275: How about just not mentioning `--force` at all and also hiding it from the --help output.

PR #375, opened 2021-08-30
Lint .travis.yml

abitrolly:
Fixes warnings from https://config.travis-ci.com/explore

[warn] on jobs.include: deprecated key: "sudo" (The key `sudo` has no effect anymore.)
[warn] on jobs.include: deprecated key: "sudo" (The key `sudo` has no effect anymore.)
[warn] on root: deprecated key: "sudo" (The key `sudo` has no effect anymore.)
[info] on root: the key matrix is an alias for jobs, using jobs
[info] on env: the key matrix is an alias for jobs, using jobs
[info] on root: missing os, using the default "linux"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: python
+os: linux
+dist: xenial
 
-sudo: false
 env:
-  matrix:
+  jobs:
     - DOCS="true"
     - TESTS="true"
   global:
@@ -14,16 +15,12 @@ python:
   - 3.5
   - 3.6
 
-matrix:
+jobs:
   include:
     - python: 3.7
-      dist: xenial
-      sudo: true
       env:
         - DOCS="true"
     - python: 3.7
-      dist: xenial
-      sudo: true
       env:
         - TESTS="true"
 


asmeurer:
FYI, ever since Travis effectively ended all free jobs, we no longer have Travis running on this repo, and it's very difficult to test any changes to doctr.

abitrolly:
I see https://docs.travis-ci.com/user/billing-overview/#free-plan - so the credits are already over? Have you tried to contact the support to give you OSS excemption?

asmeurer:
I haven't. I may try, but unless they also give me an exception for other projects I was using Travis for, I personally won't have any investment in this project any more, at least insofar as Travis support goes (if we add support for other CI(s) that may change).

abitrolly:
Understood. With a GitHub Actions killing other CI systems, I guess there is not choice anymore than to switch to the platform proposition.

abitrolly:
It also makes it easier to push docs to other branch. This is how I push a huge static build of fedora packages info https://github.com/abitrolly/fedora-packages-static/blob/8787cbfb0dd0ed957996592465751c2bf7a110ce/.github/workflows/manual.yml#L46-L68

asmeurer:
I don't think GitHub Actions is to blame for Travis stopping their free tier.

abitrolly:
But what is the point in using Travis for OSS projects right now? It is so inconvenient, and probably more costly not for OSS as well.

